PR #13, opened 2017-10-20
Update Add-SPDistributedCacheServiceInstance.md

Techwriter40:


kenwith:
It looks like the build is complaining because -Role and -CacheSizeInMB parameters were removed but they can be seen by customer in product. The tool used to generate those is called PlatyPS and it can see whatever customer can see. If they are seen, but removed, this issue will show up.

PR #67, opened 2017-11-28
update toc to add index.md

v-savila:
for bug 1132957

### DIFF
--- a/sharepoint/docs-conceptual/toc.yml
+++ b/sharepoint/docs-conceptual/toc.yml
@@ -1,3 +1,5 @@
+- name: Home
+  href: index.md
 - name: Overview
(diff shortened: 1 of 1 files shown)


v-savila:
@Techwriter40, I've added index.md per work request from Samantha Roberton (see task at https://mseng.visualstudio.com/CSI/_queries?id=1132957) and have updated toc.yml in the conceptual content folder so that the TOC will appear in the landing page (see the mockup in task) as well, however, I wanted to sync with you to confirm if we should add index.md or should it replace overview,md.

kenwith:
This looks good to me. I will sync with @Techwriter40 (Kirk) and if any issues we can alter it. Thanks!

PR #70, opened 2017-12-01
Update Set-PnPWebPartProperty.md

alicoskun85:


kenwith:
Thank you for catching the Get / Set keyword difference and thanks for contribution!

PR #76, opened 2017-12-03
Update Set-SPDiagnosticConfig.md

fowl2:
note that LogCutInterval is in minutes


kenwith:
Thanks for contribution @fowl2 !

PR #77, opened 2017-12-04
Update Set-CsUser.md

Cloud-Writer:
Request for this updated from Support based on RFC: https://icm.ad.msft.net/imp/v3/incidents/details/52553560/home

Justification: The Section is confusing and does not clearly state what number format is allowed or how. Following RFC to PG got confirmation that we only support plain E164 i.e. TEL:+14255551297 or E164+Ext in format TEL:+14255551297;ext=51297 where ext= must be part of E164 Number.


kenwith:
The error with the SPOTenantSyncClientRestriction that popped up is a different issue that will be resolved shortly. This looks good, merged! Thanks for contribution!

PR #81, opened 2017-12-05
Update Add-TeamUser.md

hmank:
On the page where it states: In this article on the right-hand menu there are 4 links for Input pointing to the same link.

### DIFF
--- a/teams/teams-ps/teams/Add-TeamUser.md
+++ b/teams/teams-ps/teams/Add-TeamUser.md
@@ -6,7 +6,7 @@ title: Add-TeamUser
 online version: 
 schema: 2.0.0
 ---
-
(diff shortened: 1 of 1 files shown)


kenwith:
The issue with the multiple links is a result of how docs.microsoft.com renders the heading 3. Since each input is a heading 3 it renders it as 3 links. I think I fixed this in this article: https://raw.githubusercontent.com/MicrosoftDocs/office-docs-powershell/live/teams/teams-ps/teams/Set-TeamFunSettings.md and you can do something similar here. It looks like this Pull Request just adds a text though and doesn't fix the issue. Please fix issue and resubmit. Thanks.

kenwith:
Closing this one as I saw you fixed in #86, Thanks!

PR #88, opened 2017-12-08
Update Install-CsMirrorDatabase.md

hmank:
The file path should be \\atl-fs-001.litware.com\DbBackup" not "\\\\atl-fs-001\DbBackup"


kenwith:
Thank you for your contribution to make the content better @hmank !!

PR #96, opened 2017-12-13
Update Set-CsClientPolicy

digitaltorres:
The -AddressBookAvailability parameter does not apply to Skype for Business Online. As indicated earlier in the article (see https://docs.microsoft.com/en-us/powershell/module/skype/set-csclientpolicy?view=skype-ps#description) and through error messages when attempting to run on 365 tenant. Error reads: 'you are not permitted to invoke set-csclientpolicy with the following parameters: addressbookavailability'. Ran as a GA.


kenwith:
Thank you for your contribution @digitaltorres  !! It is greatly appreciated!

PR #120, opened 2018-01-15
Update Get-TeamMessagingSettings.md

expiscornovus:
Removed the second empty example. It doesn't have any code.


kenwith:
Thank you for the fix!

PR #121, opened 2018-01-15
Update Install-SPFeature.md

diedvh91:
Fix -CompatibilityLevel description


kenwith:
Good catch, thank you for update!

PR #122, opened 2018-01-15
Update Install-SPFeature.md

diedvh91:
Small fix to -CompatibilityLevel example


kenwith:
Thank you for this catch and the update!

PR #124, opened 2018-01-19
Update Search-CsClsLogging.md

shanehoey:
fixed typo on -EndTime parameter


kenwith:
Thank you for the catch and making the effort to fix @shanehoey ! It is greatly appreciated!

PR #11, opened 2017-10-12
Correct format on related links. ref #10

tomlarse:
Ref #10 

This was fairly easy with Powershell, I have no clue how to do it in Python :) Just ran 

```
foreach ($file in (Get-ChildItem -path .\*.md)) {
    Set-Content $file.name ((Get-Content $file.name) -replace "\[(\w*-\w*)\]\(\)", '[$1]($1.md)')
}
```

As far as I can tell every file is fixed.

Issue #10, opened 2017-10-12
Links in Skype cmdlets related links sections are wrong


tomlarse:
If you browse a cmdlet at https://docs.microsoft.com/en-us/powershell/module/skype/?view=skype-ps, the links in the "Related links" section all seem to be linking back to the page itself.

For instance on https://docs.microsoft.com/en-us/powershell/module/skype/new-csserverapplication?view=skype-ps, the related links are to Set-CsServerApplication and Remove-CsServerApplication, but both links point back to https://docs.microsoft.com/en-us/powershell/module/skype/new-csserverapplication?view=skype-ps instead. 

In the [.md file](https://github.com/MicrosoftDocs/office-docs-powershell/blame/master/skype/skype-ps/skype/Get-CsConversationHistoryConfiguration.md#L124-L126) it seems to be right with 

```
[Remove-CsConversationHistoryConfiguration]()

[Set-CsConversationHistoryConfiguration]()
```

This seems to be the case on all (or at least many, everyone I've checked so far) cmdlets

kenwith:
Confirmed as a bug and I opened an issue with the team that handles the push from GitHub to docs.microsoft.com. Thank you for reporting! I will leave this open until I can get a resolution.

kenwith:
Just a quick update. The format needs to be like this:

[Get-CsDeviceUpdateRule](Get-CsDeviceUpdateRule.md)

I will need to write a program to manipulate these since there are so many. Or if anyone else wants to volunteer to do it. :)   I am thinking Python .. plan is to get this done and updated within a couple of weeks. I will update this issue when done.

kenwith:
OMG I LOVE YOU! :)  Merged and links should be working shortly!

thedanfernandez:
This is great thanks for the contribution!

tomlarse:
Glad to be of any help :)

PR #129, opened 2018-01-23
Fixed 'related links' links. Ref #10 and #11

tomlarse:
@Techwriter40 

Here you go Kirk. Only 266 of the files actually had any related links under "Related links", so only those were changed. 

ref #10 and #11


kenwith:
Thank you for this @tomlarse ! You rock!

tomlarse:
Glad to be of any help :)

PR #130, opened 2018-01-23
Fixed more 'related links'

tomlarse:
@kenwith @Techwriter40 

Looked through the other two subfolders in sharepoint-ps, and there were some more files that needed the fix.


tomlarse:
it seems nothing was really changed in sharepoint-pnp. I'll close and only do the sharepoint-online ones

PR #276, opened 2018-03-11
Update Dismount-SPSiteMapDatabase.md

tseward:
Filled in place holder text.


Techwriter40:
Trevor, this cmdlet is not in the product, so I think it sufficient just to say "do not use". If we have a description the customers **may** think it is OK to use.   Thoughts??

tseward:
Updated to Do not use..

kenwith:
@tseward  - looks like it is complaining that it wants a heading.. maybe move the Do Not Use to below the heading and see if that resolves it. Thanks for submitting all the PRs!

PR #312, opened 2018-03-11
Update Install-SPWebPartPack.md

tseward:
Fix example. Fill in placeholder text.


Techwriter40:
@kenwith any idea why the conflict and how to resolve it?  I looked at it, and can't see why the conflict msg is appearing.  thx

PR #316, opened 2018-03-11
Update Move-SPAppManagementData.md

tseward:
Fix example text formatting.


Techwriter40:
Trevor, 

why do you have the PS command prompt syntax in the description section? it should be straight text, "do not use". let me know. thanks.

tseward:
Was thinking there was a required consistent format, so a comment made sense in my mind. I'm happy to change it.

Techwriter40:
if you could please change it. for consistency, PS C;> and ##Comments only should be in the EXAMPLES section. thanks!!

PR #630, opened 2018-04-07
Update Set-SPExcelServiceApplication.md

tseward:
Updated example. Filled in most placeholder text.


tseward:
@Techwriter40 need info from the PG on the usage of -WorkbookModelUsageTracker.

PR #679, opened 2018-04-12
Master to Live.

kenwith:
Removes Get and Set for CsInteropPolicy cmdlet since they are no longer in product.


kenwith:
@chrisda - I noticed some major exchange so wanted to confirm with you to make sure you ok to push this master to live.

Issue #985, opened 2018-05-10
Running on SharePoint Online


superglue777:
Description states that it can be run on SharePoint Online, but seems this is not true. Power Shell does not recognise this command from SharePoint Online management shell. Are there any Power Shell plugins required to do that?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cf861e60-d88b-0a5d-2321-efdb0e281d8f
* Version Independent ID: 901da99f-c80d-743f-e0ff-f9c53bd2922a
* Content: [Register-SPAppPrincipal (sharepoint-server)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-server/register-spappprincipal?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-server/Register-SPAppPrincipal.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-server/Register-SPAppPrincipal.md)
* Service: **sharepoint-2016-powershell**
* GitHub Login: @techwriter40
* Microsoft Alias: **kirks**

Techwriter40:
thank you for your submission.  I have removed references to SPO and republishing.

PR #988, opened 2018-05-10
Update Register-SPAppPrincipal.md

tseward:
Address #985.


kenwith:
@tseward - looks like an error? Can you resolve? Let me know if I can help.

PR #1064, opened 2018-05-20
Update example

jruckert:
Fixed the example to reference the correct user claim property


kenwith:
@jruckert thanks for catch and making the docs better! Sorry for the long delay.

PR #1136, opened 2018-05-30
Update Convert-SPWebApplication.md

StefanFeenstra:
missing " in parameter


kenwith:
@StefanFeenstra - thanks for the catch and for contributing! Much appreciated.

PR #1150, opened 2018-05-31
Update Set-Mailbox.md

Cloud-Writer:
The parameters "MessageCopyForSendOnBehalfOfEnabled" and "MessageCopyForSentAsEnabled" currently do not specify that they are available on for Shared Mailboxes, misleading customers and engineers when they're trying to manipulate this behavior.


Cloud-Writer:
Don't know how/why the pull request I just created closed on its own.

PR #1208, opened 2018-06-08
Fixed two little typos

robdy:
Also tested that `Get-Module --listavailable` doesn't work so it has to be changed

### DIFF
--- a/repo_docs/NEW_CMDLETS.md
+++ b/repo_docs/NEW_CMDLETS.md
@@ -7,9 +7,9 @@
     
     **HINT:** Use `Upgrade-Module` and also `Uninstall-Module` depending on the module version you have installed.
 1. Use PlatyPS to generate Markdown.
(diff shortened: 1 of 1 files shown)


robdy:
@kenwith not sure why but for me (PowerShell 5.1) it doesn't work when I use two dashes and it works if I use one.

ISE:
2018-06-08 23_07_30-windows powershell ise

powershell.exe:
2018-06-08 23_07_49-c__windows_system32_windowspowershell_v1 0_powershell exe

Unless something has changed the convention is to use one dash as parameter prefix ([here](https://github.com/PowerShell/PowerShell/issues/1584) GitHub issue regarding this from last year).

kenwith:
@robdy - Thanks for catch. I think I have spent to much time in Linux lately. I did just try and I don't think double dash ever worked for me so I thin I am just losing it on my memory. :) Thanks for PR and fix!

PR #1219, opened 2018-06-10
Added hint of equivalent SfBO cmdlet for Get-CsUser

robdy:
I recently realized that I was not the only one confused with `Get-CsUser` and `Get-CsOnlineUser` (first used for on-premises users, second for the same but in SfB Online) - example [here](https://stackoverflow.com/questions/50765207/get-csuser-error-powershell). I think that'd be useful to include this as a hint.


kenwith:
Thanks!

PR #1239, opened 2018-06-14
Patch 5

dariomws306:
I see -DomainController parameter in Set-CsHybridApplicationEndpoint and Get-CsHybridApplicationEndpoint but not in New-CsHybridApplicationEndpoint and Remove-CsHybridApplicationEndpoint. They are SfB Server cmdlets so I think it should be useful if all of them have this parameter.
I don't understand why we have Grant-CsIPPhonePolicy if we don't have the cmdlet New-CsIPPhonePolicy to create new policies, so we don't need Grant- because Global Policy will apply to all users.
Why some SfBO cmdlets have -DomainController parameter? I leave it with "This parameter is reserved for internal Microsoft use.".
I see with PlatyPS that Grant-CsExternalUserCommunicationPolicy have the parameter -Global but I could see it originally in Github.
I see the parameter -BypassDualWrite in some cmdlets but not with PlatyPS and SfBO PowerShell.
I see that SfB Online cmdlets -CsIPPhonePolicy and -CsExternalUserCommunicationPolicy doesn't have parameter type so I have added them.

### DIFF
--- a/skype/skype-ps/skype/Get-CsExternalUserCommunicationPolicy.md
+++ b/skype/skype-ps/skype/Get-CsExternalUserCommunicationPolicy.md
@@ -14,7 +14,7 @@ Returns information about one or more external user communication policies confi
 ## SYNTAX
 
 ```
-Get-CsExternalUserCommunicationPolicy [[-Identity] <Object>] [-BypassDualWrite <Object>] [-Filter <Object>] [-LocalStore] [-Tenant <Object>] [-AsJob] [<CommonParameters>]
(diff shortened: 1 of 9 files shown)


kenwith:
@tomkau and @islubin - here is next batch of cmdlet content. Let me know if we can merge or if there are any edits. @dariomws306 added some questions and comments in the PR.

tomkau:
I'll ask the team why we have Grant-CsIPPhonePolicy, but no New.

kenwith:
@tomkau - should I go ahead and merge? We can change later too but if the content is good it will be good to get it merged.

tomkau:
Yes, go ahead and merge

PR #1285, opened 2018-06-19
Update Set-CsConferencingPolicy.md

MilleHH:
When -MaxVideoConferenceResolution is set to CIF or VGA, the Skype for Business client and Skype for Business 2015 AVMCU offers H.264


kenwith:
@tomkau - can you confirm this PR? @MilleHH is support escalations engineer and they are seeing H264 being offered in the client logs and AVMCU but we just want to confirm with you before we merge. Thanks!

kenwith:
@tomkau - any word? I am pretty confident based on what escalations engineers are seeing.

tomkau:
Yes, let's add the note.  There are references to H264 in the code as well.

PR #1296, opened 2018-06-20
Unable to use some of parameters in SfBO

naminakao:
[The following parameters are not applicable to Skype for Business Online:] section has the follwoing parameters.

-AllowCustomerExperienceImprovementProgram
-AllowExchangeConnectivity
-AllowSaveCallLogs
-EncryptAppData
-RequireIntune
-VoiceSettings

However description of them says the parameter applies to Skype for Business Online.
We confirmed those parameters did not work with SfB Online.  So I deleted SfB Online from "applies to" section.

### DIFF
--- a/skype/skype-ps/skype/Set-CsMobilityPolicy.md
+++ b/skype/skype-ps/skype/Set-CsMobilityPolicy.md
@@ -375,7 +375,7 @@ When set to True (the default value) mobile users will be allowed to participate
 Type: Object
 Parameter Sets: (All)
 Aliases: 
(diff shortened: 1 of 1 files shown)


kenwith:
hi @naminakao - thank you for the PR. This is a known issue. The reason the applicable tag is showing up for SfBO is that these parameters are showing up when PlatyPS runs. The PlatyPS tool just sees whatever is available and reports it as available for the cmdlet. So if we remove the applicable like in PR then PlatyPS will just add it back and see it as a 'new' parameter next time it runs. We are looking into separating which cmdlets show for SfBO vs. SfB Server.

PR #1344, opened 2018-06-26
Update to ComplianceRetentionPolicy to state a location parameter is required to create a retention policy

Tangela123:
To create a ComplianceRetentionPolicy the cmdlet requires at least one location (exchange, sharepoint, etc) to be defined.


chrisda:
@golinku I fixed this in the main branch.

PR #1379, opened 2018-07-02
StartTime should be last 30 days

naminakao:
I cannot get data if I set the date more than 30 days ago as StartTime. Seems I can get data for  last 30 days. I found the following report.
If we changes the design to provide user session date only for last 30 days, please mention that in public content. 

-GDPR: Set retention to 30 days for get-csusersession related Baja tables​​​​​​​
https://skype.visualstudio.com/SBS/_workitems/edit/1319303

### DIFF
--- a/skype/skype-ps/skype/Get-CsUserSession.md
+++ b/skype/skype-ps/skype/Get-CsUserSession.md
@@ -36,6 +36,7 @@ This example returns user session information for Ken Myer from "02/22/2016 07:3
 
 ### -StartTime
 PARAMVALUE: DateTimeOffset
+StartTime should be last 30 days.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


kenwith:
@naminakao - can you let us know about the comment being added regarding the last 30 days? As far as we can understand the only requirement is the start time is before the end time.

kenwith:
@naminakao - we can't see why the start time should be last 30 days. We looked in code and only validation is that it must be earlier than end time. Going to close this one out but please re-open if you have more info. Thanks for taking time to submit and make the docs better!

kenwith:
I think this was covered in another PR. Thank you @naminakao for submitting! If the other PR didn't resolve please re-open.

PR #1476, opened 2018-07-19
Update Set-CsTeamsMeetingBroadcastConfiguration.md

tomkau:
Added external help file reference


islubin:
@tomkau is this just for the purpose of the new XML drop or is there something else I need to be doing when updating these docs?

tomkau:
This is for generating the XML.  It gave a warning that the file reference was missing and did not include it in the XML.

PR #1491, opened 2018-07-24
Update Set-CsMediaConfiguration.md

bament-hun:
updated with latest platyPS output

### DIFF
--- a/skype/skype-ps/skype/Set-CsMediaConfiguration.md
+++ b/skype/skype-ps/skype/Set-CsMediaConfiguration.md
@@ -1,7 +1,7 @@
 ---
-external help file: Microsoft.Rtc.Management.dll-help.xml
-applicable: Lync Server 2010, Lync Server 2013, Skype for Business Server 2015, Skype for Business Server 2019
-title: Set-CsMediaConfiguration
+external help file: Microsoft.Rtc.Management.dll-Help.xml
+Module Name: SkypeforBusiness
+online version:
 schema: 2.0.0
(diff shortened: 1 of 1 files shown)


skype/skype-ps/skype/Set-CsMediaConfiguration.md
```diff
@@ -256,64 +235,45 @@ Accept pipeline input: False
...
 
 Required: False
 Position: Named
 Default value: None
-Accept pipeline input: False
+Accept pipeline input: True (ByValue)
 Accept wildcard characters: False
 ```
 
-### -EnableH264Codec
```
tomkau: I still see this parameter in the code branch I have access to.  

@kenwith, can you ask the Server 2019 team if this is being removed in 2019?  Even so, it seem we should keep the parameter in the docs for older version.

skype/skype-ps/skype/Set-CsMediaConfiguration.md
```diff
@@ -256,64 +235,45 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
-### -EnableG722StereoCodec
```
tomkau: Same comment as the H264 codec.

skype/skype-ps/skype/Set-CsMediaConfiguration.md
kenwith: Yes, will start an email thread with the PM and include you both.

PR #1494, opened 2018-07-25
Update New-TeamChannel.md

hrfmjp:
can't contain the characters was wrong for display.  I removed unnessesary line break.


kenwith:
@hrfmjp - thank you for your contribution to make the docs better! Much appreciated!

PR #1497, opened 2018-07-26
disable PDF for testing

EdisonDa:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
qinezh: no need to check in this file?

PR #1513, opened 2018-07-28
Update New-Team.md

revaldiv:
Added: " Note: The Teams application may need to be open by an Owner for up to two hours before changes are reflected. " at Line 15

Need to clarify that Owner must have client open and adding additional hour to provisioning time.


islubin:
Approved.

PR #1521, opened 2018-07-31
Doc changes to make modern authentication easier

bspender:
The first 6 or 7 examples rely on legacy authentication. I would request moving those options lower (to deprioritize their usage) and to move the modern authentication options higher. Plus, in the text, I added more verbiage to help folks get past the errors received when legacy authentication has been disabled ...and a nudge towards the options that do support modern auth.


kenwith:
@bspender - this makes a lot of sense and thanks for taking the time to make the docs better! I just have to figure out the status errors showing here before we can merge.

kenwith:
@bspender - I finally figured it out.. the schema was complaining about the 3 back ticks.. I changed to 1 back tick for inline code and seems to work. Merging now. Thanks!

PR #1542, opened 2018-08-05
Dariomws306 patch 3

dariomws306:
I don't know what -ShareData parameter do on Set-CsOnlineVoicemailPolicy and what is the -User parameter type on Get-CsUserSession.
-SiteName and -MediationServerIPAddress are string on -CsHybridPSTNAppliance cmdlets?

Thanks!


qinezh:
Reopened the PR to trigger a new build.

kenwith:
@tomkau - OK to merge? I also saw @dariomws306 had a question at top.

tomkau:
@dariomws306 

The -User parameter is required to specify which user's session data is being requested.

Yes, -SiteName and -MediationServerIPAddress are both strings.

For -ShareData, I found the following comment in code:

Specifies whether (voicemail and transcription) data is shared with the service for training and improving accuracy. By default, it is "Defer" which means that it is upto the user to decide whether they want to share their voicemail data with the service team (configurable through User Settings Portal). Other possible value is "Deny" which means that the voicemail data for any user in the tenant will not be shared with the service regardless of the users own choice.

dariomws306:
Thanks @tomkau !!

About -User my question was not clear. There I have set type as UserUri, is that ok? I am not sure about type.

Thanks!

tomkau:
Ah, I see.  This cmdlet has -User as just a string.

PR #1548, opened 2018-08-07
Update .openpublishing.publish.config.json

Banani-Rath:


qinezh:
It would be better to use the release version instead of prerelease version for stability.

kenwith:
@Banani-Rath and @qinezh - I am not sure I follow...  any thoughts @dstrome and @chrisda ?

dstrome:
@qinezh, is this related to https://icm.ad.msft.net/imp/v3/incidents/details/79189446/home or to the Exchange moniker changes we requested? Agree we shouldn't be using pre-release plugins but need someone from APEX engineering to determine if this change is ok.

cc: @Banani-Rath @EdisonDa

PR #1552, opened 2018-08-07
Update Get-CsUserSession.md

dariomws306:
Only fix -User


dariomws306:
@tomkau @kenwith could we merge? I just fixed -User parameter type.

kenwith:
Yes, merged. Thanks!

PR #1570, opened 2018-08-10
Patch 10

dariomws306:


dariomws306:
@tomkau could you review please?

I am not sure about -AdditionalAudienceUrls and -AlternateAudienceUrl on Set-CsOAuthConfiguration.

Thanks!

tomkau:
@dariomws306, you can use the following description for both audience params:

An OAuth token includes an audience ("aud") claim that identifies the intended recipient of the token, in this case your Skype for Business Server.  The audience is typically expressed as a URL, such as https://skypeforbusiness.contoso.com, and is validated by the Skype for Business authentication module.  In some cases, depending on your authentication providers and/or other applications used in your organization, it may be necessary to allow multiple audience values to be used and recognized as valid by your Skype for Business server.

The AlternateAudienceUrl and AdditionalAudienceUrls parameters allow you to specify one or more additional audience values that will be recognized as valid by your Skype for Business server.

tomkau:
@dariomws306, from what I can tell, there should not be much need for ClientAdalAuthOverride anymore. Let's use the following in the docs.  I'm intentionally being vague about ASDF/MEX to discourage non-advanced admins from doing it.

Modern authentication using Azure ADAL (Active Directory Authentication Library) is enabled by default for Skype for Business Online.   For authentication based on ADFS/MEX (metadata exchange), ADAL can be disabled by setting ClientAdalAuthOverride to Disallowed.

dariomws306:
@tomkau thanks!

And what about the type of parameter? It's Object now.

tomkau:
The fully qualified object type is:
Microsoft.Rtc.Management.WritableConfig.Settings.SSAuth.Hosted.ClientAdalAuthOverride

dariomws306:
@kenwith I think we are done here.

PR #1579, opened 2018-08-14
AppDesktopSharingMode parameter cannot set with New-Cs​Teams​Meeting​…

naminakao:
…Policy

When we tried to set the parameter with New-Cs​Teams​Meeting​Policy, the error "A parameter cannot be found that matches parameter name 'AppDesktopSharingMode'." occurred.

PR #1580, opened 2018-08-14
CloudRecordingServiceSupport cannot set with Set-CsConferencingPolicy

naminakao:
I requested content update when this page was on public in TechNet, Content Idea Request 64449.
When the content moves to doc.microsoft.com and updated part seems reverted. 

Based on following bug, it seems the parameter, -CloudRecordingServiceSupport has been removed from Get-CsConferencingPolicy and no longer works.
  Bug: Office15 3491382 - CloudRecordingServiceSupport parameter in LYO Conference Policy

When we set -CloudRecordingServiceSupport parameter with  Set-CsConferencingPolicy command, we don't get any error but cannot get the value of this parameter from running Get-CsConferencingPolicy.


kenwith:
@tomkau - OK to merge?

tomkau:
I don't see anything in the code to suggest this param is not supported.  It might be a bug that it is not being returned.  Let me investigate.

kenwith:
See comment here: https://github.com/MicrosoftDocs/office-docs-powershell/pull/1579  regarding PlatyPS and if the param shows up in code.

tomkau:
I've confirmed with the team that this is a cmdlet bug-- the value is being properly set, but just not returned.  I've opened bug to track.

@naminakao, please use this description for the parameter:
Specifies at conference creation and activation time whether to allow the organizer to select use of the Cloud Recording Service or to always force/disallow use.

kenwith:
@naminakao - thanks for catching this! If I understand correctly the cmdlet has a bug and will match the reference when fixed. I will close this out. Thanks for your effort on these. Much appreciated!

PR #1689, opened 2018-09-04
Patch 9

dariomws306:
@kenwith @tomkau 

I don't know what "-RouteType" parameter do on New-CsOnlineVoiceRoutingPolicy and Set-CsOnlineVoiceRoutingPolicy.
Could you help me with this? Or should I use "This parameter is reserved for internal Microsoft use"?

Also, I am thinking that I should have write more information about that these cmdlets are related to Teams Direct Routing. What do you think?

Thanks!

### DIFF
--- a/skype/skype-ps/skype/Get-CsOnlinePstnUsage.md
+++ b/skype/skype-ps/skype/Get-CsOnlinePstnUsage.md
@@ -23,7 +23,9 @@ Get-CsOnlinePstnUsage [-Tenant <Guid>] [-Filter <String>] [-LocalStore] [<Common
 ```
 
 ## DESCRIPTION
-PSTN usages are string values that are used for call authorization. A PSTN usage links a voice policy to a route. The `Get-CsOnlinePstnUsage` cmdlet retrieves the list of all PSTN usages available within a tenant.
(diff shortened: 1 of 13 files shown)


kenwith:
@tomkau - OK to merge?

tomkau:
@dariomws306 -- from what I can see in code, the parameter -RouteType is not used.  Let's use your suggestion of saying reserved for internal Microsoft use.

There is already good documentation for direct routing, so we do not need to add a lot of details about direct routing in the cmdlet documentation.  But, for the relevant cmdlets, it's good to note that they are used to configure direct routing.  Admins can then look up more details as needed.

For the relevant cmdlets, we can add something simple like:

"This cmdlet is used when configuring Microsoft Phone System Direct Routing."  

I'm not very familiar with direct routing and how to configure, but if you read this article it should help you identify which are used with dircet routing configuration.
https://docs.microsoft.com/en-us/MicrosoftTeams/direct-routing-configure

dariomws306:
@tomkau thanks!!

I have been using the DR article you sent me and onpremises cmdlets as referece because they are very similar.

I have applied all the comments/details you sent me.

dariomws306:
@kenwith @tomkau do you think we are done with this?

tomkau:
Yes, let's go ahead and push these.

PR #1723, opened 2018-09-12
Remove some arguments from CSMeetingConfiguration

TomonoriAbe:
From bellow bug, EnableAssignedConferenceType and AssignedConferenceTypeByDefault are not supported by MeetingConfiguration. So please remove SPO from these argument to prevent confusion of admins.
https://skype.visualstudio.com/SBS/_workitems/edit/1468690

### DIFF
--- a/skype/skype-ps/skype/Set-CsMeetingConfiguration.md
+++ b/skype/skype-ps/skype/Set-CsMeetingConfiguration.md
@@ -167,7 +167,7 @@ The default value is True.
 Type: Boolean
 Parameter Sets: (All)
 Aliases: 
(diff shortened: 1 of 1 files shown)


kenwith:
@TomonoriAbe - thanks for this PR and for taking the time to make the content better.

This is a known issue. You are correct that technically the applicable should be changed but unfortunately those parameters still show up in the code so if we remove the 'applicable' then when we run the open source PlatyPS tool (https://github.com/PowerShell/platyPS) it will 'see' those parameters and list them as 'new' parameters. We are working on a system for this so stay tuned.

@islubin and @tomkau  - FYI.

PR #1800, opened 2018-10-02
Merges Master to Live for a number of commits.

kenwith:


chrisda:
2 hours is too long; I'll fire up a new one.

PR #1809, opened 2018-10-03
Update Set-OrganizationConfig

JuForan:
Document the new parameter that we're going to be rolling out soon.

It's not yet live for everybody, but I assume that's OK?


chrisda:
New PR coming. Error were caused by incomplete YAML section in the new parameter description. Other additions and deletion appear to be caused by an out-of-date fork.

PR #1814, opened 2018-10-04
Update New-Team.md to standardize format

sz02:
Fields are specified in camel case in other Team pages. Fix inconsistencies in all-lower-case and tick-usage here.


kenwith:
@sz02 - thank, much appreciated!

PR #1818, opened 2018-10-04

Cloud-Writer:
I have added the following to the description for CustomFooterText: Multiple lines are supported using the Enter key. This request came via Content Idea 92143. Thanks!


kenwith:
@Cloud-Writer Thanks Meera!

PR #1825, opened 2018-10-05
remove duplicate line

ttplaya:
Removed duplicate bullet in the list of minimum requirements under the Permissions section


kenwith:
@ttplaya - thanks for catch and fix, much appreciated!

PR #1833, opened 2018-10-09
Suggested Updates to New-PnPTenantSite.md

pkbullock:
The LCID references do not list the ID codes for creating a site with a specific LocaleID with a type of UINT32 so have added a command to get that from SharePoint online. Minor corrections to the other commands and seemingly corrected a bad URL at the bottom of the article.

### DIFF
--- a/sharepoint/sharepoint-ps/sharepoint-pnp/New-PnPTenantSite.md
+++ b/sharepoint/sharepoint-ps/sharepoint-pnp/New-PnPTenantSite.md
@@ -75,7 +75,7 @@ Accept pipeline input: False
 ```
 
 ### -Lcid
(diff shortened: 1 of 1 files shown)


pkbullock:
Feel free to reject, just seen the notice on the homepage readme. If someone contributes via the edit button on docs site they would not see the warning, as I have just done.

kenwith:
@pkbullock - thanks for contribution, much appreciated! There is a check that looks at the PowerShell schema ( https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md ). It is super picky and never very clear what causes an error. I have even had just a single character complain on me and it has taken me forever to figure it out. :)  

Also, thanks for pointing out the warning in whiteboard content. I will check that out and get that resolved. Feel free to re-open or ping me if you want to go ahead with a change to the PnP cmdlets.

PR #1839, opened 2018-10-10
Add run guidance.

kenwith:


kenwith:
@Fogolan - I added some guidance in README. Would you please check and let me know if OK to merge it? Make sure all developers do a 'pull' to get latest too once merged.

PR #1870, opened 2018-10-12
Update New-Team.md

18h:
both Get-Team and Set-Team pages provide links to New-Team, so it'd be nice to provide direct links from the other direction too


kenwith:
@18h - thanks for taking time to make docs better. Much appreciated!

Issue #508, opened 2018-03-28
Set-SPOTenant -UserVoiceForFeedbackEnabled:$false seems to have no effect on modern pages


wobba:
I ran `Set-SPOTenant -UserVoiceForFeedbackEnabled:$false` on two tenants and nothing changes (checked after 24h).

I would expect the "Feedback" link in the lower right hand corner to disappear when setting this value to false.

![image](https://user-images.githubusercontent.com/51104/38026245-b7e67da0-328b-11e8-8151-1cb1b947e6ad.png)

kenwith:
I will check with @VesaJuvonen who we can check with on Product Team. I would expect the Feedback to disappear too.

VesaJuvonen:
I'm now looking for the internal owner so that we can make this fixed. It clearly seems to be a bug on the modern sites for ignoring this setting. Will come back asap.

wobba:
@VesaJuvonen feel free to alert the owner of the "Get the mobile app" link as well - wouldn't hurt to hide it.

inorde:
@VesaJuvonen any news?

KatjaJokisalo:
I'm following this as well. Changing the value doesn't seem to have any effect.

VesaJuvonen:
Just pinged again internal threads around this so that we could get this moving. Stay tuned.

VesaJuvonen:
Fix for this is currently rolling out worldwide and should be completely out within week or two. Thanks for reporting and let's close this issue when we have full confirmation that the fix is properly working.

gurubn:
Is this setting (Set-SPOTenant -UserVoiceForFeedbackEnabled:$false) expected to hide Feedback button in delve as well ?

wobba:
I can confirm this is working in several tenants on modern pages. @gurubn, it will only work on modern pages right now, not system pages, like Delve.

ColinGardner:
It work for me as well but as with others, only on standard pages.  Can we also have a way to hide the "Get the mobile app" button as this only addresses half the issue.  It shows an incredible lack of foresight not to have this sorted when modern was released let alone to still not have a proper solution many months down the line, when clearly upwards of 80% of people don't want this feature.  And whilst we are at it, please can we have a proper web UI setting for this instead of having to resort to PowerShell for everything.

clkmsc:
[...]

PR #1885, opened 2018-10-16
Provided explanation for UserVoiceForFeedbackEnabled

KoenZomers:
With the fix to the -UserVoiceForFeedbackEnabled flag as announced here https://github.com/MicrosoftDocs/office-docs-powershell/issues/508#issuecomment-430152475 it now also takes care of removing the "Get the mobile app" link. This argument didn't have a comment with it yet to explain what it does, so added this.


kenwith:
@KoenZomers - thanks, appreciated!

PR #1891, opened 2018-10-17
Update Get-PnPTenant.md

expiscornovus:
Added another example to get the tenant settings, but this time sorted ascending. 

Probably is a better way doing this, but this is just a workaround ;)


kenwith:
@expiscornovus - much appreciated!

PR #1936, opened 2018-10-21
remove Skype for Business Server 2019 from all the Applies To - issue 1928

maryfranceshull:


kenwith:
@maryfranceshull - I thought I got all of these but I missed this one. Thanks for catch and fix! Much appreciated!

PR #1952, opened 2018-10-23
Update Remove-MessageClassification.md

tsahi:
remove redundant header; remove empty Notes header; remove Related Links section which redirects back to the same page.

### DIFF
--- a/exchange/exchange-ps/exchange/policy-and-compliance/Remove-MessageClassification.md
+++ b/exchange/exchange-ps/exchange/policy-and-compliance/Remove-MessageClassification.md
@@ -27,8 +27,7 @@ You need to be assigned permissions before you can run this cmdlet. Although thi
 
 ## EXAMPLES
 
-### -------------------------- Example 1 --------------------------
-```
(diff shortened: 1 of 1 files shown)


tsahi:
Am I to understand that the execution of the PowerShell snippet failed?

kenwith:
@tsahi - yes, the automatic checks failed. The checks are really picky since the references need to follow the exact schema for PowerShell since the references go back into the product so Get-Help <cmdlet> works correctly. The schema can be found here: https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md  

It can be tricky figuring out what is wrong. I can try to help too if you have questions. Thanks for making the docs better!

kenwith:
@tsahi - I added the Example 1 header back in since I think the schema might have been wanting it.. let's see if it will pass

kenwith:
@tsahi - one note, the related link for the Online Version is required so that when you are in command line and do   Get-Help -online   then the page will open to the correct location. Please add that back in and we should be good. Thanks!

PR #1963, opened 2018-10-24
typo: pubic -> public

nschonni:


kenwith:
@nschonni - wow, bad typo! Thanks for fix!

PR #1964, opened 2018-10-24

YuriySamorodov:
I was pointed by Microsoft Support Engineer lately that since SharePoint is not listed under Applicable field, this cmdlet is not intended to be used in order to retrieve SharePoint Online  logs and this does not make a lot of sense.


chrisda:
@YuriySamorodov: sorry, man. The Applicable tag is a documentation thing for the applicable versions of Exchange PowerShell for the cmdlet, not the Office products that it works with. Therefore, we need to leave the value at Exchange Online.

YuriySamorodov:
@chrisda No worries. I expected that. If possible, just let your support team know that Exchange Online stands for Powershell modules required by this cmdlet rather than Exchange Online service itself.

kenwith:
As Chris mentioned, the applicable is for the tool that builds the help into Get-Help so we can't really change it. Appreciate the info on this though!

PR #1986, opened 2018-10-27
Fixed StartDate and EndDate parameters

YuriySamorodov:


chrisda:
@YuriySamorodov the lack of square brackets around StartDate and EndDate indicates these parameters are required. Putting square brackets around the parameters indicates they're optional. Was that the intent?

PR #1988, opened 2018-10-27
Update README.md

MAHIMA01:


kenwith:
@MAHIMA01 - thanks for taking the time to review and make the docs better! To add a break  in markdown you can add a second carriage return instead of the HTML <br/>.  Please make update and resubmit. Thanks!

PR #1989, opened 2018-10-27
Update LICENSE

praveen-x:


kenwith:
@praveen-x - thanks for the effort on this to make the LICENSE better. 

@AndreaBarr - for any of these files we will need to get a sign off from CELA (legal) before even a simple grammar can be fixed.

kenwith:
@praveen-x - it looks like the license is correct as it is, here is the link: https://creativecommons.org/licenses/by/4.0/legalcode  going to close this one out but please re-open if I am missing something. Thanks very much for taking the time to make the docs better too. It is much appreciated.

PR #1998, opened 2018-10-29
Fixed example text for "-SkipServer"

greiginsydney:
Hostnames need to be fully qualified.


kenwith:
@greiginsydney - thanks for fix and making the docs better. Much appreciated!

PR #2021, opened 2018-11-02
Small fix in wording

Szab:
#### Category
- [x] Content fix
- [ ] New article

#### What's in this Pull Request?
Changed _Remove an owner or member from a team, and **to** the unified group which backs the team._ to _Remove an owner or member from a team, and **from** the unified group which backs the team._


kenwith:
@Szab - thanks for fix, much appreciated!

PR #2033, opened 2018-11-03
Update New-MigrationBatch.md

dimcry:
Updated the old Technet articles, to the new Docs ones.


chrisda:
@dimcry I'm closing the PR because a) the existing redirects from TechNet to docs.microsoft.com work fined (I checked them all) and b) you need to strip down the Docs links a bit first; don't include the locale (en-us) and don't add the extra text after the cmdlet name (? and everything to the right).

PR #2146, opened 2018-11-18
Typo fixed

Apoc70:
changed $fase to $false


kenwith:
@Apoc70 - thanks for catch and fix! Much appreciated.

PR #2156, opened 2018-11-20
typo (missing verb)

bb-froggy:


kenwith:
@bb-froggy - thanks for catch and fix, much appreciated!

PR #2176, opened 2018-11-26
Update set-spotenant

Vridhi-Chowdhry:
There is a typo in the command. Corrected here.


kenwith:
@Vridhi-Chowdhry - thanks for catch and fix. Much appreciated!

PR #2178, opened 2018-11-26
Update Set-PnPSite.md

expiscornovus:
Small typo in code example 5. Should be LogoFilePath instead of SiteLogoPath.


kenwith:
@expiscornovus - thanks for catch and fix, much appreciated!

PR #2190, opened 2018-11-29
Docfix for DisabledWebPartIds in Set-SPOTenant

KoenZomers:
The documentation on the new DisabledWebPartIds argument seems inaccurate. It mentions using Get-SPOSite and Set-SPOSite while I believe it should be Get-SPOTenant and SPO-Tenant instead. Corrected that in this PR. Also clarified how you would be able to enable all webparts again by providing DisabledWebPartIds $null.


KoenZomers:
During my tests against my tenant I noticed that running Set-SPOTenant -DisabledWebPartIds $null doesn't actually bring back all previously blocked WebParts. So that part of my submission does not seem to be correct. Trying to figure out what the correct way of enabling them all again would be.

kenwith:
@VesaJuvonen - I think this relates with https://github.com/MicrosoftDocs/office-docs-powershell/pull/2219  OK to merge?

KoenZomers:
@kenwith, correct my change has been done in discussion with @YusongLiu. Would be best to merge both our PRs together as they both contain a few different details.

kenwith:
@KoenZomers - I think I merged together correctly and will push live as soon as the checks finish here. Thanks for the catch on these and making the content better, much appreciated! If I missed anything or made any mistake please feel free to update.

PR #2210, opened 2018-11-30
Minor typos + changed -whatif description

greiginsydney:
For consistency's sake I changed the -whatif description to the same string used by other articles in this section.


kenwith:
@greiginsydney - thanks!

PR #2217, opened 2018-12-02
Mispelling

dmadelung:
Mispelling


kenwith:
@dmadelung - thanks for catch and fix!

PR #2227, opened 2018-12-05
Updating sample for Remove-SPOTenantCdnOrigin

sottenad:
the -OriginScope parameter is no longer accepted. It looks like it was removed from the documentation but not from the sample. Closes issue #2225


kenwith:
@sottenad - thanks for fix, much appreciated!

PR #2240, opened 2018-12-07
Update Set-IRMConfiguration

SSalahAhmed:
Removed DecryptAttachmentFromPortal because it is deprecated.
Updated ClientAccessServerEnabled as it is no longer honored and will be deprecated.


kenwith:
@SSalahAhmed - thanks for your contributions and making the docs better. We should only remove the actual parameter if it has been removed from the code. If it is deprecated for a specific version it is worth putting a note for it. Looks like there are some errors with the PR. I would recommend adding the note on what is deprecated but leaving the parameters. I will close this one out since there are errors in the checks and please feel free to add a new one. Thanks, taking the time to make the docs better for the community is greatly appreciated!

PR #2262, opened 2018-12-12
Update Set-CsTenantFederationConfiguration.md

robertpp:
small change


kenwith:
@robertpp - thanks for catch and fix!

PR #2299, opened 2018-12-17
Corrected typos, renamed to "CS-" references to "Cs-"

greiginsydney:


kenwith:
@greiginsydney - thanks for these fixes and making the content better. Love how it is constantly improving now with experts like you sharing your knowledge. Much appreciated!

Issue #2306, opened 2018-12-17
Update examples


NicholasBurkhammer:
Example 1 and Example 2 both have Set-CsAuthConfig followed by a parameter called -BlockWindowsAuthExternally or -BlockWindowsAuthExternallyAndModernAuthInternally.  These items are not parameters.  They are values for parameter -Scenario.  Both should be updated to have Set-CsAuthConfig -Scenario in front of the actual parameter value which should not have a hyphen in front of it.

Example:  Set-CsAuthConfig -Scenario BlockWindowsAuthExternally

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 26656471-f4e4-3dbe-b915-5230400d7b7f
* Version Independent ID: 4dc006ee-6bf8-a929-94f7-499a5af1f7a1
* Content: [Set-CsAuthConfig (skype)](https://docs.microsoft.com/en-us/powershell/module/skype/set-csauthconfig?view=skype-ps)
* Content Source: [skype/skype-ps/skype/Set-CsAuthConfig.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/skype/skype-ps/skype/Set-CsAuthConfig.md)
* Service: **skypeforbusiness-powershell**
* GitHub Login: @kenwith
* Microsoft Alias: **kenwith**

kenwith:
@NicholasBurkhammer - looks like @nenonix (Nicole) submitted a Pull Request to update this for you. Just a note though that you can also update articles yourself. You can check the README on the main repo (https://github.com/MicrosoftDocs/office-docs-powershell) to see how to do it. This way you can fix articles right away instead of opening an issue for someone else to do it. I am happy to guide you too when you want to make an update. Thanks.

PR #2308, opened 2018-12-17
Update Set-CsAuthConfig.md

nenonix:
Edits made to example one and two (added -Scenario and removed hyphen from parameter) as per https://github.com/MicrosoftDocs/office-docs-powershell/issues/2306


kenwith:
OPS Build status updates of commit _[038e6bd](https://github.com/nenonix/office-docs-powershell/pull/2308/commits/038e6bdbccdee5ad0ecf64720f36406f61554b95)_: 

:clock10: **Full build**: average full build time is _11 min(s) 54 sec(s)_ [calculated from last _1_ build(s)]

> Note: A full build takes more time than an incremental build.  This is the expected behavior of the build system.

PR #2395, opened 2018-12-28
Update Disable-UMAutoAttendant.md

SignorelliDenis:
There is no UM commands in Exchange Server 2019 because the UM was discontinued.


chrisda:
@SignorelliDenis, true, but the UM cmdlets are supposed to still be there for coexistence with Exchange 2013 and Exchange 2016. I went through all the trouble of what you just did in the 50+ UM cmdlet ref topcs, but then I had to add the information back in right before 2019 RTM.

I know during the 2019 Preview, the UM cmdlets weren't available. They were supposed to be added back in time for RTM.

SignorelliDenis:
@chrisda Actually, I’m talking about RTM version (Build 15.2.221.12). In my lab those UM commands are not exposed. I tried two different labs, one with pure Exchange 2019 RTM and other with coexistence between 2016 and 2019. So as you told I supposed that those commands were not added in time for the RTM.

PR #2419, opened 2019-01-04
Update Set-AntiPhishPolicy.md

dfantham:
Removing parameters that are not applicable or available WW yet


kenwith:
@dfantham - thanks for making the content better, much appreciated!

@chrisda - OK to merge or should we leave the parameters and add a note they are not yet available? Thoughts?

PR #2454, opened 2019-01-09
added new HardDelete to PD for PurgeType

markjjo:


sujithnair79:
PurgeType Harddelete is only available in Exchange Online. This is not available in Onpremise Exchange 2016 and 2019 at all.

This is help section from the onpremise Exchange 2016 server.

[PS] C:\Windows\system32>get-help New-ComplianceSearchAction -Parameter Purgetype
 
-PurgeType <Unknown | SoftDelete>

     The PurgeType parameter specifies how to remove items when the action is Purge.
 
    The valid value for this parameter is SoftDelete, which means the purged items are recoverable by users until the deleted items retention period expires.
 
    Required?                    false
     Position?                    Named
     Default value
     Accept pipeline input?       False
     Accept wildcard characters?  false

PR #2497, opened 2019-01-15
Updating Get-SPOTenant output properties

bcameron1231:
Addressing issue #2473 where the Get-SPOTenant command does not display which output parameters are available from the PowerShell command. 

A couple values are missing because I was unable to find information on what they mean.


kenwith:
@bcameron1231 - the cmdlet content is unique in that it has to follow a strict schema.. I have spent hours and hours over the last couple of years trying to track down errors like you are experiencing above so an help find them. Here is the schema the check uses: https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md  it complains if the headings aren't exact and a lot of other little things.

kenwith:
One thing is that we can't use tables in the cmdlet reference.. since this is the same content that shows up in the terminal if someone runs    Get-Help <cmdlet>

bcameron1231:
@kenwith  Thank you for the feedback. Is there a preferred layout for this type of content? Or should we just include documentation on the get-help? I'd happy to PR a better approach.

kenwith:
@bcameron1231 - sorry, I wasn't clear. The   Get-Help   command will show what is in this article you are working on. There is an import process. So the Get-Help output is this article you are working on.

You can put the information in this cmdlet reference, it just has to follow the schema. I think the schema for outputs involves a header with a description underneath (instead of a table). 

Here is an example of a cmdlet reference that has a significant amount of info in it: https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/skype/skype-ps/skype/New-CsClientPolicy.md 

You can see the raw Markdown if you click on the Raw button at the top of the page.

Thanks for this work, it is greatly appreciated and community really appreciates it! Let me know any questions or if you get stuck and I can help.

bcameron1231:
Ah okay. Great. Let me see if I can some up with something using that supports this layout.

kenwith:
@bcameron1231 - going to close this for now so branch doesn't get too far out of date. Ping me back when you want to work on it and we can work together to get info in, I think it is really valuable and greatly appreciate your effort on this!

PR #2499, opened 2019-01-15
Cannot use displayname as identity

naminakao:
When I use display name value for identity parameter when I run set-addresslist in ExO, it fails with error. Error said "Object 'display name' not found in   'KAWPR01A001DC02.JPNPR01A001.PROD.OUTLOOK.COM'".


kenwith:
@naminakao - I noticed you are really active in the docs and it is greatly appreciated! Just wondering if you could share if you are part of Microsoft?

PR #2503, opened 2019-01-16
Update Set-HostedContentFilterPolicy.md

ryancairnes:
Hi KC & Denise, can you please review these changes?


kenwith:
@ryancairnes - thanks for updates and making the docs better, much appreciated!

Looks like you are looking for a review from @KCCross and @denisebmsft 

Also, @chrisda FYI

PR #2517, opened 2019-01-17
Update Get-CsOnlineApplicationEndpoint.md

dariomws306:
I found "The following parameters are not applicable to Skype for Business Online: AsJob, Audience, CallbackUri, DomainController, Force, IsInternalRun, PipelineVariable, Ring, and RunFullProvisioningFlow" here:
https://docs.microsoft.com/en-us/powershell/module/skype/set-csonlineapplicationendpoint?view=skype-ps
https://docs.microsoft.com/en-us/powershell/module/skype/new-csonlineapplicationendpoint?view=skype-ps

### DIFF
--- a/skype/skype-ps/skype/Get-CsOnlineApplicationEndpoint.md
+++ b/skype/skype-ps/skype/Get-CsOnlineApplicationEndpoint.md
@@ -8,32 +8,31 @@ schema: 2.0.0
 # Get-CsOnlineApplicationEndpoint
 
 ## SYNOPSIS
-Provide the topic introduction here.
+It is used to fetch the application endpoints for the tenants.
 
 ## SYNTAX
 
 ```
-Get-CsOnlineApplicationEndpoint [-Uri] <String> [-Audience <String>] [-Ring <String>] [-PhoneNumber <String>]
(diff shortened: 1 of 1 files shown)


kenwith:
@dariomws306 - this is a known issue and specific to PowerShell. Here is what is going on... the parameters are in the code for some reason.. so the tool that automatically stubs out the content, using reflection, sees those parameters and puts them in. The tool is open source and called [PlatyPS](https://github.com/PowerShell/platyPS).

Currently the product teams have been going through and manually fixing the PowerShell references to make things match what 'should' be correct. However, this causes a 'delta' between what is in the reference content and what is in the code. 

We have a new tool that is very close to rolling out (it is open source and in the tools directory in this repo) and it will update the cmdlet reference content using reflection. So if it sees a parameter was removed from code it will remove it, if it sees a new parameter or cmdlet was added it will stub out the content. So very soon any parameters (like in this PR) that were removed for correctness will re-appear again as 'new' parameters until they are removed from the code. This will force product teams to actually remove parameters that are not needed or should not be used as well.

I know that is a long answer but that where we currently at. I will go ahead and merge this one but just wanted to give that answer so if these parameters show up as 'new' in the near future everyone knows what is going on.

PR #2534, opened 2019-01-19
chore: Remove package-lock.json

nschonni:
There is a lock file, but no package.json


kenwith:
@nschonni - thanks for catch! This was a remnant from the tool being created and slipped past me when I merged. Much appreciated!

PR #2680, opened 2019-01-26

dariomws306:
EnableAssignedConferenceType is not available in SfB Online.


kenwith:
@tomkau - looks correct that this parameter was removed in SfBO?

tomkau:
@kenwith, yea, the code does not allow this parameter for Online.  Seems to have been that for a few years now, but the docs never got updated...

PR #2733, opened 2019-02-04
Update Restore-RecoverableItems to include PurgedItems switch

Lakshminarayand9:
Update Restore-RecoverableItems to include PurgedItems switch with SourceFolder


Lakshminarayand9:
as per the article Restore-RecoverableItems, It only support recovering messages from DeletedItems or RecoverableItems. 
However when I tested I noticed we can also use Purgeditems as a source folder parameter

PR #2734, opened 2019-02-04
Update Get-RecoverableItems to included Purgeditems parameter

Lakshminarayand9:
Please included Purgeditems parameter with source folder switch


Lakshminarayand9:
As per the article GeT-RecoverableItems,  to view deleted items in the Recoverable Items folder  amd deleted items in mailboxes, However when I tested I noticed we can also use Purgeditems as a source folder parameter

PR #2768, opened 2019-02-08
Fix headers and revert duplicate files for LBR related cmdlets

kelqi:


skype/skype-ps/skype/Get-CsTenantNetworkRegion.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Get-CsTenantNetworkRegion
+
+## SYNOPSIS
+
+## SYNTAX
+
+```
+Get-CsTenantNetworkRegion [-LocalStore] [[-Identity] <Object>] [-Tenant <Object>] [-Filter <Object>] [-AsJob]
+```
```
tomkau: There's an quirk with PlatyPS in that it sets all parameters to be Objects, when in fact some are strings or numerical.  I can't remember if this is fixed in a later version.  Ken would know.

Meanwhile, you should double check the cmdlet source code to see if the parameter types are correct, and update as needed.

skype/skype-ps/skype/Get-CsTenantNetworkRegion.md
```diff
@@ -0,0 +1,116 @@
...
+
+## DESCRIPTION
+{{Fill in the Description}}
```
tomkau: You need to add descriptions, examples, etc.

kelqi: Engineers will add once it's checked in

skype/skype-ps/skype/Get-CsTenantNetworkRegion.md
kelqi: New PR with fix. Need to add -Session param when running platyPS command to avoid this issue

kelqi:
New PR with correct variable types in params.

PR #2791, opened 2019-02-12

msbemba:
made the change for the attributte -DesignateAsPresenter   Company by default and not to none.


kenwith:
@tomkau and @islubin - does this change look correct? Seems weird to me that PlatyPS set to "None"...

kenwith:
@yogkumgit - I am just trying to get approval from @tomkau or @islubin on this change to make sure it is something they are aware and agree.

tomkau:
The change is correct--matches the code

kenwith:
Thanks @tomkau ! Will merge.

Issue #2634, opened 2019-01-21
Conflicting information for parameters applicable to Skype for Business Online


rgsteele:
The last paragraph of the Description reads "The following parameters are not applicable to Skype for Business Online: AsJob, CalendarPollInterval, Force, Identity, Instance, PipelineVariable, SIPSecurityMode, Tenant, Voice8021p, and VoiceDiffServTag". However, within the description for each of these parameters, "Skype for Business Online" appears in the "Applies to" row of the table (except for AsJob and PipelineVariable, which don't appear at all).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 27501e53-7915-e774-a5f8-153a86e4b497
* Version Independent ID: 826e461a-0bb8-eb12-aff0-4f27831f5f71
* Content: [Set-CsUCPhoneConfiguration (skype)](https://docs.microsoft.com/en-us/powershell/module/skype/set-csucphoneconfiguration?view=skype-ps)
* Content Source: [skype/skype-ps/skype/Set-CsUCPhoneConfiguration.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/skype/skype-ps/skype/Set-CsUCPhoneConfiguration.md)
* Service: **skypeforbusiness-powershell**
* GitHub Login: @kenwith
* Microsoft Alias: **kenwith**

get-itips:
Hi @rgsteele Thanks for your feedback! Actually, that parameters are present in Online Version so I created a PR to remove that from description. Thank you

yogkumgit:
@kenwith Please merge this PR if you find it suitable

kenwith:
This is a known issue. The 'applies to' is auto generated by a tool called PlatyPS that uses reflection to see what cmdlets and parameters are available from a user point of view. PlatyPS then spits out the cmdlet stubs and sets the applies to. Some parameters are legacy from Skype for Business Server so they appear in Skype for Business Online even though they are not used. This issue will be resolved in the near future but for now we added that sentence at the top of the description since the 'applies to' is more of 'cmdlet contains' for the parameters. We thought of adding a 'do not use for Skype for Business Online' to all of the relevant parameters but there are many of them and this issue is across many cmdlets. And also this will be fixed. 

[...]

PR #2891, opened 2019-02-26
Removed comment about params not present in Online

get-itips:
Removed comment about parameters not being present in Online version https://github.com/MicrosoftDocs/office-docs-powershell/issues/2634


kenwith:
Not positive on this one, @islubin - I remember we added this statement.. should it be removed now if something changed or should it be left? It is a 'known issue' regarding cmdlets that appear in SfBO but are actually only for SfB Server.

kenwith:
Going to close this for now. See description in the associated Issue. This is a constant source of confusion and we are working on it. FYI @islubin @tomkau

PR #2935, opened 2019-03-01
Update Set-SPOSite.md

msbemba:
Added parameter set 4 with restrictions on a groups site collection.


msbemba:
creating a new one after signing CLA

PR #2973, opened 2019-03-05
Dariomws12

dariomws306:
-ExternalEmailAddress appeared 2 times.


chrisda:
1: Don't remove duplicate YAML blocks from parameter descriptions. There's always a reason why there is more than one. In this case, the ExternalEmailAddress parameter was required in the DisabledUser and EnabledUser parameter sets, and optional in MicrosoftOnlineServicesID and FederatedUser parameter sets. If you've tested and found this isn't true, you need to add or remove the square brackets from around the parameter in the Syntax block(s) at the top of the topic, and also move the parameters up or down in the list in the Syntax block(s) (required=near the top; optional=alpha within the rest). And, please comment in the pull request that you've tested this and absolutely confirmed the topic is currently wrong.

2: The first link in the Related Links section is what makes `Get-Help <Cmdlet> -Online` work correctly for all Exchange cmdlets. If you find the link doesn't work correctly, replace it with the docs.microsoft.com equivalent. I tested https://technet.microsoft.com/library/128467a7-b8b8-4fa6-bca9-1131301f18ce.aspx and it correctly redirected to https://docs.microsoft.com/powershell/module/exchange/users-and-groups/New-MailUser. Beware of transient issues with redirection. If a link that's formatted correctly.

chrisda:
It should also say that parameters with multiple YAML blocks are currently being duplicated on docs.microsoft.com. I have a bug in to see what's going on (I think it's a recent change, but I can't say for sure).

dariomws306:
Thank you for this information!

PR #3026, opened 2019-03-07
Update Set-CsUser.md

carolynblanding:
Please prioritize this. Only 3 of the parameters on this page apply to Skype for Business Online and one of these 3, OnPremLineURI, only applies to Skype for Business online. 

I hope the changes I made clarify this. 

Big concern here is that it appears that customers can configure ExchangeArchivingPolicy for online users and that is not correct. Currently, all SfB Online users are configured with a Uninitialized policy that cannot be modified by Tenant Administrators which means that SfB content will be archived if the users mailbox is placed on hold. 

We are seeing thousands of errors a week for customers with on premises mailbox and online SfB users where archiving back to Exchange server on premise is failing. They likely won't be aware of this and if they go to look for how to manage this policy will be steered in the wrong direction. 

Please let me know if you have any questions or contact hars@microsoft.com regarding priority of this work! Thanks.


kenwith:
@carolynblanding - the <sup> tags won't render in the Get-Help in the PowerShell terminal so we need to figure out a way to make it clear without using those tags. Maybe break the syntax blocks into those supported with Online or not? I can help on this too.

kenwith:
@yogkumgit - would you please see if a freelancer can help? I cannot merge this PR since the tag is not supported. Please ask a freelancer to review this PR and create a new one.

PR #3040, opened 2019-03-10
fix: MD025/single-h1

nschonni:
Multiple top level headings in the same document

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Short URL: aka.ms/office-powershell
 
 ## Overview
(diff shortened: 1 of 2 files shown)


kenwith:
The checks seem stuck. I will check back on this later and ping the team that owns the build check system if this doesn't clear.

nschonni:
Feel free to try and just redo it on your internal fork if this is just jammed

kenwith:
I have seen a thread with other repos experiencing this too so it is something with the build process. I will keep checking that thread and see what they resolve then get this one through. Thanks!

PR #3050, opened 2019-03-10
fix: CsTeamsUpgradePolicy

nschonni:
- adminstrators -> administrators
- adminsitrators -> administrators
- remove raw GitHub links

### DIFF
--- a/skype/skype-ps/skype/Set-CsTeamsUpgradePolicy.md
+++ b/skype/skype-ps/skype/Set-CsTeamsUpgradePolicy.md
@@ -13,35 +13,39 @@ ms.reviewer:
 # Set-CsTeamsUpgradePolicy
 
 ## SYNOPSIS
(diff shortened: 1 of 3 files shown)


skype/skype-ps/skype/Set-CsTeamsUpgradePolicy.md
```diff
@@ -132,18 +135,18 @@ For more information, see about_CommonParameters (http://go.microsoft.com/fwlink
 
 ## RELATED LINKS
 
-[Migration and interoperability guidance for organizations using Teams together with Skype for Business](https://docs.microsoft.com/en-us/MicrosoftTeams/migration-interop-guidance-for-teams-with-skype)
+- [Migration and interoperability guidance for organizations using Teams together with Skype for Business](https://docs.microsoft.com/en-us/MicrosoftTeams/migration-interop-guidance-for-teams-with-skype)
```
nschonni: Need to undo the list formatting, used in other document formats, but the PowerShell processor doesn't like it under the RELATED LINKS header

kenwith:
We use an open source tool called PlatyPS to stub out and generate the markdown reference from cmdlet modules using reflection. Here is the schema it uses: https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md

There could definitely be a bug in what PlatyPS is doing. If so we can open on that project and hope it gets worked on.

PR #3054, opened 2019-03-11
typo: extentions -> extensions

nschonni:
- additioanlly -> additionally
- ingetration -> integration
- Trim trailing spaces
- Add code fence langauges
- Space areound headers
- Space around code fences
- Malformed link

### DIFF
--- a/exchange/exchange-ps/exchange/client-access-servers/Set-OwaVirtualDirectory.md
+++ b/exchange/exchange-ps/exchange/client-access-servers/Set-OwaVirtualDirectory.md
@@ -12,7 +12,6 @@ monikerRange: "exchserver-ps-2010 || exchserver-ps-2013 || exchserver-ps-2016 ||
 # Set-OwaVirtualDirectory
 
 ## SYNOPSIS
-
 This cmdlet is available only in on-premises Exchange.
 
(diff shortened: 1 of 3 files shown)


exchange/exchange-ps/exchange/client-access-servers/Set-OwaVirtualDirectory.md
```diff
@@ -2880,10 +2880,12 @@ This cmdlet supports the common parameters: -Debug, -ErrorAction, -ErrorVariable
 
 ## INPUTS
 
+###
```
nschonni: @chrisda  I was wondering about these since I saw them in a few files. I see that the build fails if they're missing, but an empty header is also not valid for Markdown.

chrisda:
@nschonni: Don't feel obligated to "fix" perceived markdown issues in Exchange/Exchange Online/EOP/Security & Compliance Center cmdlet topics. I'd rather have them all be the same kind of wrong than have a handful of them fixed. For example:

- No need to make web links hot; they won't be hot at the command line, and the important URL part will disappear if you format them "correctly" per markdown.

- No need to put "powershell" in front of code blocks.

- The INPUTS and OUTPUTS sections require the ### subsection to hold the text. When you removed them, it caused build errors.

- Don't worry about missing carriage returns under section headings.

Thanks.

chrisda:
And now I see there's a build error in the SharePoint file. I'm closing this PR

PR #3055, opened 2019-03-11
fix: MD039/no-space-in-links

nschonni:
- Spaces inside link text
- Trim trailing spaces
- Adjust headings


kenwith:
We use an open source tool called PlatyPS to stub out and generate the markdown reference from cmdlet modules using reflection. Here is the schema it uses: https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md

There could definitely be a bug in what PlatyPS is doing. If so we can open on that project and hope it gets worked on.

PR #3090, opened 2019-03-18
-LinkType EventSubscribers has been deprecated in Add-UnifiedGroupLinks

MSFT-Shivsi:
EventSubscribers from lines 25, 114 & 121 need to be removed. It has been deprecated. Please refer to VSTS: 1047396. 
Or else please Skype IM me on Shivsi@microsoft.com


kenwith:
@MSFT-Shivsi - do you know if these were removed from code or are still in code but should not be used? If they are still in code then they will just show up as 'new' parameter next time the PlatyPS tool is run. PlatyPS runs reflection to look for new or removed parameters and sees whatever customers see.

kenwith:
If still in code but not used I think a better approach is to add a note that they are deprecated and should not be used.

PR #3150, opened 2019-03-28
Move cmdlets out from subfolders and adding group map

qinezh:
This PR does below things:
1. Move cmdlets out from subfolders under exchange/exchange-ps/exchange.
2. Add group mapping to achieve similar toc.


dstrome:
The changes in this PR are not feasible - we need to maintain the folder structure for topic management. Do not merge this PR.

jmazzotta:
@dstrome i noticed you closed Qinen's PR without merging. lets chat further as unfortunately this is the only current viable approach to render the exchange module page.

PR #3160, opened 2019-03-29
Fix parameter type in NGC AA Cmdlet document

CHILIU-MSFT:
Fix parameter type in NGC AA Cmdlet document


CHILIU-MSFT:
@kenwith Could you help approve? Thanks!

Issue #3138, opened 2019-03-26
Set-RetentionCompliancePolicy multiple (10) parameters with help text placeholders


msjoat:
Set-RetentionCompliancePolicy (office-docs-powershell/exchange/exchange-ps/exchange/policy-and-compliance-retention/Set-RetentionCompliancePolicy.md) has 10 parameters with help text placeholders, e.g. 

{{Fill AddDynamicScopeLocation Description}}

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws306:
Hi @msjoat, thank you for your feedback.

I have removed 2 parameters, they doesn't exist anymore. I am trying to get more information about the others 8.

yogkumgit:
@dariomws306  Thank you very much for the contribution and updating the documentation with PR. @msjoat Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

yogkumgit:
@officedocsbot close

PR #3210, opened 2019-04-12
Dariomws21

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/3138


chrisda:
@dariomws306, when you remove parameters, be sure to search **everywhere** in the topic. The parameter sections are the most obvious, but the parameter will always be in one or more of the parameter sets in the SYNTAX section at the top of the topic. Although it's unlikely, the could even be in one or more examples in the EXAMPLES section.

dariomws306:
> @dariomws306, when you remove parameters, be sure to search **everywhere** in the topic. The parameter sections are the most obvious, but the parameter will always be in one or more of the parameter sets in the SYNTAX section at the top of the topic. Although it's unlikely, the could even be in one or more examples in the EXAMPLES section.

@chrisda yes I know that sorry I forgot it in this case :( .

Thank you for your help!

PR #3228, opened 2019-04-16
Regards `DuplicateOutputFiles` as build error to avoid 404

qinezh:


kenwith:
@qinezh - seems like some errors? What should we do with this one?  CC @dstrome @chrisda

PR #3231, opened 2019-04-17
Disable incremental build

qinezh:
Currently there's a known issue for PowerShell reference build pipeline, which may cause 404. This PR is to disable incremental build temporary before the issue is resolved.


kenwith:
@qinezh - you know way more than me about this one. CC to @dstrome and @chrisda  I am fine with any change that needs to be made.

PR #3360, opened 2019-05-06
Corrected private key requirement

ykuijs:
The page incorrectly stated that a private key had to be present. However it should NOT be used, just like with New-SPTrustedRootAuthority. See: https://docs.microsoft.com/en-us/powershell/module/sharepoint-server/new-sptrustedrootauthority?view=sharepoint-ps


tseward:
This correction does not make much sense given the PFX cmdlet included requires that the private key for the certificate. See https://docs.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Security/Get-PfxCertificate?view=powershell-5.0. And the PFX certificate format must have a private key.

In addition, the SharePoint Root Authority must have a private key (you can't trust the cert w/o it).

ykuijs:
The root certificate that SharePoint generates for the local farm indeed requires the private key to be present in order for it to work. However the Set-SPTrustedRootAuthority doesn't manage that certificate, but adds the root certificate of other farms to the store of the local farm, so a trust is created between two farms. In that case, importing the private key would introduce a security issue since the second farms private key is no longer private. That is why the cmdlet throws an error when the specified certificate contains a private key. In [SharePointDsc](https://github.com/PowerShell/SharePointDsc/pull/687/files) we specifically added some code to strip out any private key if a certifcate was specified that contains a private key, to prevent the module from running into an exception.

Also, as mentioned in the PR description, the [New-SPTrustedRootAuthority](https://docs.microsoft.com/en-us/powershell/module/sharepoint-server/new-sptrustedrootauthority?view=sharepoint-ps) mentioned the exact opposite as this cmdlet:

> The New-SPTrustedRootAuthority cmdlet creates a trusted root authority. If a certificate file is used, it must have only one X509 certificate **without** private keys, otherwise an exception is raised.

The same is true for [Get-SPTrustedRootAuthority](https://docs.microsoft.com/en-us/powershell/module/sharepoint-server/get-sptrustedrootauthority?view=sharepoint-ps):

> The Get-SPTrustedRootAuthority cmdlet returns a trusted root authority. If a certificate file is used, it must have only one X509 certificate without private keys, otherwise an exception is raised.

In order words: The information in this page is incorrect and should state that the specified certificate has to be a certificate **without** private key.

And now that I think about it: In all examples you can find on the Internet, the specified certificate is a CER file and not a PFX file (this cmdlet doesn't have a password parameter, so how can it import a PFX without the password). And since CER files cannot contain private keys, the example provided probably also needs updating.

tseward:
Not enough coffee in the morning :) You'll want to edit the example and resubmit the PR, then.

ykuijs:
Ok, all done! Ready for review/merge!

ykuijs:
Quick question: When will this be merged?

ykuijs:
When will this PR be merged?

ykuijs:
@trent-green Any clue when this PR will be merged? It has been updated with the master branch several times already, but isn't merged afterwards.

ykuijs:
@chrisda @islubin I see that you have merged PRs in the past. This PR is ready for merge for weeks now, but so far hasn't been merged. Could one of you please merge this PR?

PR #3403, opened 2019-05-13
Update Enable-SPOCommSite.md

tseward:
Removing en-us/gb from links; fix link.


kenwith:
@trent-green @tseward - seems like something in metadata isn't rendering right. Check if there is a space after the ms.reviewer?

image

kenwith:
Ok Andres found it, looks like there is a space before the opening --- at the front of file.

kenwith:
@get-itips Andres found it I mean.

kenwith:
Ok fixed and building now.

PR #3506, opened 2019-06-04
Update Set-CsUser.md

carolynblanding:
@kenwith @bjwhalen 
Updating this to remove Skype for Business SERVER distinction since this cmdlet applies to both Server and Online. Added some context about that it update SfB Attributes in Azure AD for Enterprise Voice, Hosted Voicemail, and on-premises line URI. 

Removed Skype for Business Online reference in applicable to for all parameters where this is not true. 

Change driven by customer confusion on this cmdlet and how it's used online versus on premises AND because we need to be clear that customers cannot modify the Exchange Archiving Policy when a user is online.


bjwhalen:
That’s fine, but there are a few spots where we only list onprem versions

Shouldn’t this…

Applicable: Lync Server 2010, Lync Server 2013, Skype for Business Server 2015, Skype for Business Server 2019

Be…

Applicable: Lync Server 2010, Lync Server 2013, Skype for Business Server 2015, Skype for Business Server 2019, Skype for Business Online

Issue #1665, opened 2018-08-30
This cmdlet usage need to be clarified


mrik23:
If we use "-Role Owner" the user is only added to as Owner of the group, this seems fine with Teams but not with all workloads.

For example, if an user is only set as Owner of a Group and not a member, that user will not see the plan associated with the Office 365 group in Planner (or it is Planner that doesn't work correctly).

When we create an Office 365 Group from Teams, Outlook or SharePoint, we are added as both member and owner, so I guess it is the right way.

Maybe you should make Add-TeamUser to add the user as member and owner, or you should clarify the documentation so we first add the user as member and then as owner.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: da0c53f2-1c62-14f6-6371-f54bf86b0c24
* Version Independent ID: aac162db-a9c0-fbfb-dd40-526e94d3bef8
* Content: [Add-TeamUser (teams)](https://docs.microsoft.com/en-us/powershell/module/teams/add-teamuser?view=teams-ps)
* Content Source: [teams/teams-ps/teams/Add-TeamUser.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/teams/teams-ps/teams/Add-TeamUser.md)
* Service: **teams-powershell**
* GitHub Login: @kenwith
* Microsoft Alias: **kenwith**

AndreaBarr:
@mrik23 Thank you for this feedback

I will get this over to the PowerShell team so we can get this article updated. 

I was able to get this into the backlog for the PowerShell team. The team will get this prioritized into a sprint plan. Thank you for reporting and making the docs better. Much appreciated. 
 
I made a note to request the team to update this when the work is complete.

dariomws306:
Hi @mrik23 , thank you for your feedback.

I am adding a line to let readers know that if they use this cmdlet to add an owner, they need to run this cmdlet again to add him as member.

dariomws306:
PR:
https://github.com/MicrosoftDocs/office-docs-powershell/pull/2786

moegitree:
Agree with @mrik23. But I found the explanation part added by @adriomws306 is missed now (5/30). 
I can see from article history that it was overwritten in Apr 24.
Teams PowerShell Module 1.0 was released in the same day. But cmdlet's add owner movement remained the same as commented by @mrik23.

Sorry but could you confirm this case one more time?

[...]

PR #3542, opened 2019-06-11
Update Remove-TeamUser.md

AlexanderHolmeset:
Needed to specify user role to remove.  Docs also says that -Role is only used to demote user, but -role removes the user from the team, as pr cmd description at the top.


kenwith:
@AlexanderHolmeset - thanks for taking the time to make the docs better and submit fixes. It is much appreciated!

@islubin - can you let me know if ok to merge?

yogkumgit:
@dariomws306  , @get-itips and @tseward Please review this PR and give your feedback. @kenwith looks forward to your comments.

get-itips:
Hello @AlexanderHolmeset how are you?

I tested this in my lab, and I could verify that if you ran

`Remove-TeamUser -GroupId 31f1ff6c-d48c-4f8a-b2e1-abca7fd399df -User dmx@example.com -Role Owner`

it actually removes the ability for the user, to access the team, but, strange enough, if you ran

`Get-TeamUser  -GroupId 31f1ff6c-d48c-4f8a-b2e1-abca7fd399df `

you will see it still listed in the Powershell result, but as a member (as the current article says)

image

(You won't see it listed in the Teams UI though)

Maybe is some caché thing or the module is buggy, either way, it seems it is better to describe the current behavior of the cmdlet, as you suggest in this pr, so I'll accept it.

tseward:
We should see if the PG considers this a bug prior to changing the article. As the module itself has [-Role] as _optional_, the PR would change the article and the product would no longer match the documentation.

get-itips:
@tseward you are right on that. thanks

dariomws306:
There is a similar behaviour with Add-Teamuser (and different from Admin Center):
https://github.com/MicrosoftDocs/office-docs-powershell/issues/1665

I have done some testing and I found that the -Role paramater didn't exist in older version of Teams Powershell.

I am agree with Trevor, also I think it makes sense that this parameter is optional.

AlexanderHolmeset:
Here is a screenshot to better describe the error at my end. 
The difrence between my and @get-itips error, is that with me the user is removed and not demoted.
So probably some kind of bug in the module.
Makes sense as you say to make the parameter -Role optional. 
Maybe there should be a new cmdlet like this to demote/promote users:
Set-TeamUser -GroupID "XXXXXXX" -User "user1@contoso.com" -Role Owner

image

erwinbierens:
> Here is a screenshot to better describe the error at my end.
> The difrence between my and @get-itips error, is that with me the user is removed and not demoted.
> So probably some kind of bug in the module.
> Makes sense as you say to make the parameter -Role optional.
> Maybe there should be a new cmdlet like this to demote/promote users:
> Set-TeamUser -GroupID "XXXXXXX" -User "[user1@contoso.com](mailto:user1@contoso.com)" -Role Owner
> 
> image

This only applies to the owner role.. 

When you add/remove the user as a member this error is not displayed. 
image

AlexanderHolmeset:
Good catch Erwin.

kenwith:
@islubin can you comment on this one? Is this a bug and should be change the documentation?

kenwith:
Add Jamie Stark ( @nomorephones ) who took over Teams PowerShell from @islubin to review and sign off.

modpra:
Any updates on this? The PowerShell to remove users still does not work.

kenwith:
@nomorephones any update?

get-itips:
Tested this again with module version 1.0.1

Removed user that was owner on a group using the -Owner parameter, **and it was demoted to member.** This was reflected right away both in UI and Powershell Get-TeamUser cmdlet.

Removed user that was owner on a group NOT using the -Owner parameter, **and it was removed from the channel/group.** This was NOT reflected right away in Powershell Get-TeamUser cmdlet, and I am still waiting to see this reflected.

dariomws306:
Also tested with 1.0.1, my comments:

- Same as Andres using -Role parameter (if I use user UPN or UserId)
- Without -Role parameter:
  * If I use UserId same as Andres, user is not removed from the Team, not demoted to member and 
    no output
  * If I use UPN same as Erwin, I get the same error

Cmdlet documentation says that we should use UPN.

get-itips:
@tseward what do you think of our conclusions ?

tseward:
@get-itips we're still on hold for this one.

get-itips:
@AlexanderHolmeset Can you test again with module version 1.0.1 ? Thanks

AlexanderHolmeset:
@get-itips have the same result as you now
image

get-itips:
Hello @nomorephones 
Issue creator now experiences the same as we all do.
Based on that, and the current behavior of the cmdlet, do you think is necessary to update the documentation?
We really appreciate your time to review this and give us your thoughts.
Thank you

kenwith:
Merging this based on feedback from the product team and they have also added an internal work item to address the inconsistencies brought up above.

flowt-au:
This is still an issue, I think. Using v1.0.21
I have a Team with 1 existing owner. I add a new Owner: 

`Add-TeamUser -GroupId 557acb93-6e5d-4031-8def-xxxxxxxxx -User sally.student@xxxxx.nsw.edu.au -Role Owner`

and I can see Sally as an owner of the group. All good.
Then, if I attempt to demote her to Member as per docs:

`Remove-TeamUser -GroupId 557acb93-6e5d-4031-8def-xxxxxxxxx -User sally.student@xxxxx.nsw.edu.au -Role Owner`

Sally is removed from the Team altogether.
:-(

PR #3567, opened 2019-06-14
updated search-mailbox as per work item 118875008

markjjo:
removed references to data loss related to auto-expanding archiving in EXO


markjjo:
@chrisda can you please merge this PR to live?

PR #3580, opened 2019-06-19
Path of example should be just export.cmp

plantoschka:


yogkumgit:
@plantoschka Request you please update the Path as advised by Trevor and Andres. Thank you.

kenwith:
Can we make the change in another PR and I will close this one? @plantoschka must be busy so it would be good if we could capture his contribution and get it merged.

@get-itips would you be able to submit a PR that incorporates this PR and the feedback?

kenwith:
Actually, let me merge this one so @plantoschka is listed as contributor and then @get-itips would you have time to submit a new PR to make adjustment recommended above?

PR #3594, opened 2019-06-20
Update New-CsOnlinePSTNGateway.md

NMURAV:


kenwith:
@NMURAV - the 'applicable' is pulled from the module name so cannot be changed in the reference file.

kenwith:
@NMURAV please ping me regarding this update regarding the 'applicable' as it is pulled from the module name. There are some cmdlets in Skype for Business module that  affect Teams and vice versa. I realize it is not ideal since applicable is pulled from module name. @islubin was working with some others on a way to address. If you are in product team or in support ping me if you have ideas too get the applicable fixed in the code.

Issue #3637, opened 2019-06-28
Example 2 error


mdhirt:
Hi,

On my exchange 2013 server Example 2 was not working. 

I found that by adding the Name property to the Mailbox parameter  for Get-ActiveSyncDeviceStatistics resolved it for me.

I expect this is because the objects returned by Get-CASMailbox are not Mailbox objects as required by the Mailbox parameter of Get-ActiveSyncDeviceStatistics, however giving the parameter the CASMailbox object Name property is enough for the function to identify the mailbox.

$UserList = Get-CASMailbox -Filter {HasActiveSyncDevicePartnership -eq $true -and -not DisplayName -like "CAS_{*"}; Get-Mailbox $UserList | foreach {Get-ActiveSyncDeviceStatistics -Mailbox $_.Name}

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3e9d90b2-695e-59b0-92c2-ab9bcf43a78e
* Version Independent ID: 5647aff0-b7e9-a057-a63d-03b667a8c4b4
* Content: [Get-ActiveSyncDeviceStatistics (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/devices/get-activesyncdevicestatistics?view=exchange-ps#feedback)
* Content Source: [exchange/exchange-ps/exchange/devices/Get-ActiveSyncDeviceStatistics.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/devices/Get-ActiveSyncDeviceStatistics.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

mdhirt:
Just found the examples for Get-MobileDeviceStatistics which work much better and should probably be used in preference to Get_ActiveSyncDeviceStatistics because it is being depreciated.

https://docs.microsoft.com/en-us/powershell/module/exchange/devices/get-mobiledevicestatistics?view=exchange-ps

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws306:
Hi @mdhirt, thank you for your feedback.

You replied by yourself. This article has a note about that you have to use Get-MobileDeviceStatistics for Exchange 2013 or later.

Anyway, I have used your feedback to remove Exchange 2013, 2016, 2019 and online from this article.

[...]

PR #3664, opened 2019-07-05
Update Get-ActiveSyncDeviceStatistics.md

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/3637

### DIFF
--- a/exchange/exchange-ps/exchange/devices/Get-ActiveSyncDeviceStatistics.md
+++ b/exchange/exchange-ps/exchange/devices/Get-ActiveSyncDeviceStatistics.md
@@ -1,12 +1,12 @@
 ---
 external help file: Microsoft.Exchange.MediaAndDevices-Help.xml
-applicable: Exchange Server 2010, Exchange Server 2013, Exchange Server 2016, Exchange Server 2019, Exchange Online
+applicable: Exchange Server 2010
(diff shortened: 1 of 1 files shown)


chrisda:
@dariomws306, we need to be careful here. Although Exchange 2013, 2016, and 2019 are clearly supposed to use the \*-MobileDevice cmdlets instead of the *-ActiveSyncDevice cmdlets, the *-ActiveSyncDevice cmdlets are still _available_ in those newer versions of Exchange (including Exchange Online). I'm not sure what to do about it at this point.

dariomws306:
@chrisda, thank you for your help.

I was thinking in the note... If people should use Get-MobileDeviceStatistics on 2013 an later, I think it's a good idea to remove 2013 or later as supported in this article.

But I understand your point, if it works it's supported... So I'm not sure what to do either...

On the other hand, the issue author has an issue using Get-ActiveSyncDeviceStatistics... So maybe it doesn't work well on 2013 or later...

Maybe it could be a good idea to get more information from the guy who put this note.

chrisda:
Closing per discussion on Issue #3637. This cmdlet is **available** in Exchange 2010 through 2019 and Exchange Online, but really works best in Exchange 2010. I updated the existing note to say this.

PR #3688, opened 2019-07-15
Update Get-CsBackupServiceStatus.md

hirenshah1:
Added new category as RGS. Also changed the details that if no parameter is specified, all three types will be checked.


hirenshah1:
@kenwith @chrisda Please merge these changes

PR #3728, opened 2019-07-19
Move packaging cmdlet docs to correct directory

etgottli:


etgottli:
@chrisda Can you merge this for me? @kenwith is OOF

PR #3789, opened 2019-08-02
AdditionalStorageProvidersAvailable in-by-default

lusassl-msft:
See roadmap Feature ID: 52597
We are adding more online storage providers to Outlook on the web and making it on by default. This update will bring two changes: 1. More providers, including personal OneDrive, Google Drive, and Facebook. 2. We are making this in-by-default
https://www.microsoft.com/en-gb/microsoft-365/roadmap?rtc=2&filters=&searchterms=52597


lusassl-msft:
@chrisda can you check this PR?

Issue #3630, opened 2019-06-27
how do I filter Office 365 groups?  the cmdlet list doesn't include get-unifiedgroup, even though get-unifiedgroup takes a -filter parameter


anwarmahmood1:
if they aren't the same as these, then what are they?  Can you create a separate page?

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9ce35226-b40b-3c9f-a37f-585b84aa129e
* Version Independent ID: ac9953e2-769a-c115-b1c6-e6f6fc940db2
* Content: [Filterable properties for the Filter parameter](https://docs.microsoft.com/en-gb/powershell/exchange/exchange-server/recipient-filters/filter-properties?view=exchange-ps)
* Content Source: [exchange/docs-conceptual/exchange-server/recipient-filters/filter-properties.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/docs-conceptual/exchange-server/recipient-filters/filter-properties.md)
* Service: **exchange-powershell**
* Product: **exchange-server-itpro**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws306:
Hi @anwarmahmood1, thank you for your feedback.

There were a similar issue:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/1668

@chrisda, do you think we can add Get-UnifiedGroup here?
https://docs.microsoft.com/powershell/exchange/exchange-server/recipient-filters/filter-properties

Thanks!

yogkumgit:
@chrisda Please merge the PR #3826

yogkumgit:
@get-itips  and @tseward Please review the PR #3826

yogkumgit:
@dariomws306  Thank you very much for the contribution and updating the documentation with PR. @anwarmahmood1 Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

yogkumgit:
@officedocsbot close

PR #3826, opened 2019-08-12
Patch 31

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/3630

- Get-UnifiedGroup could be filtered by Database, ServerName, ExpansionServer and ExchangeVersion paremeters. But I didn't add it to the article. I think it makes no sense.

- Could we confirm next items?
   * ResourceProvisioningOptions possible values are $null and 'Team'?
   * Why Description, ExchangeObjectId and DataEncryptionPolicy weren't added to this article? I see 
   that them applies to all the recipients running Get-Recipient (DataEncryptionPolicy only to some of 
   them)
   * ResourceBehaviorOptions possible values are the next?
AllowOnlyMembersToPost
CalendarMemberReadOnly
ConnectorsEnabled
HideGroupInOutlook
NotebookForLearningCommunitiesEnabled
ReportToOriginator
SharePointReadonlyForMembers
SubscriptionEnabled
SubscribeMembersToCalendarEvents
SubscribeMembersToCalendarEventsDisabled
SubscribeNewGroupMembers
WelcomeEmailDisabled
WelcomeEmailEnabled


dariomws306:
Also, I have some suggestions:

•	I think it’s a good idea that other technical expert review this PR.
•	Also, as Chris David said, I think it’s a good idea to check all the table to add more Exchange Online parameters.

get-itips:
Work in progress:

It seems you have the correct values for ResourceBehaviorOptions and also for ResourceProvisioningOptions

get-itips:
Looks ok, it would be great if someone from product team could confirm all the possible values, but I think if that will delay, the PR will improve those articles anyway.

PR #3843, opened 2019-08-15

cversion7:
Applies to Microsoft Teams, but was not listed as applying to anything but Skype for Business Online.


kenwith:
@tomkau - will adding Teams to the "applies to" break the process of getting the reference content checked in and working with Update-Help? The PowerShell reference is technically part of the Skype for Business Online module and I know the applies to field is used when the CAB file is generated so this PR might break that. If you want to use this as a test case to give it a try I am OK with that.

Issue #3893, opened 2019-08-27
Wildcard accepted in -RecipientAddress


boaz-w:
Wildcards seem to be accepted for the RecipientAddress parameter unlike stated in the documentation.
For example, Start-HistoricalSearch -RecipientAddress "*@gmail.com" works as expected.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 38a22164-2df5-5dc2-9292-df60df018a53
* Version Independent ID: 1d96478c-230b-889e-68ed-9a8ab9484be5
* Content: [Start-HistoricalSearch (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/mail-flow/start-historicalsearch?view=exchange-ps#feedback)
* Content Source: [exchange/exchange-ps/exchange/mail-flow/Start-HistoricalSearch.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/mail-flow/Start-HistoricalSearch.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

get-itips:
Hello @boaz-w thanks for your feedback. I created a pull request to improve the article. Thank you

yogkumgit:
@chrisda Please merge the PR #3908

yogkumgit:
@get-itips  Thank you very much for the contribution and updating the documentation with PR. @boaz-w Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #3908, opened 2019-08-29
RecipientAddress works with wildcard chars

get-itips:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/3893


chrisda:
For reference, there's no programmatic way to determine the correct value of that Accepts wildcard characters? property. It's a systemic Windows PowerShell problem. Any True values you see on any cmdlets have been manually changed. Up to you re: whether to update these values or not.

get-itips:
Thanks @chrisda I too think that changing that to True is going too deep but as I verified what the user said, decided to update it. Is up to you to merge this pr or not. I trust on your call.

PR #3931, opened 2019-09-03
Cmdlet documentation for batch policy assignment

tomkau:
Two new cmdlets available in the Teams module for public preview


tomkau:
Need to fix validation errors

PR #3951, opened 2019-09-05
[Set-Mailbox] adding ability to remove delay hold applied with groupmailbox 

telaws:
I recently checked in code that allows the ability to remove a delay hold applied from a group mailbox using parameters -GroupMailbox and -RemoveDelayHoldApplied:

set-mailbox testgroupmailbox -GroupMailbox -RemoveDelayHoldApplied

Prior to my change this was not possible for a Group Mailbox.  Additionally, I don't see the documentation listing parameter -GroupMailbox. 

Please update the documentation to cover above scenario and parameters.


chrisda:
Whenever you manually add parameters to a cmdlet reference topic, you also need to add the parameter in the SYNTAX block(s) (parameter sets) at the top of the topic. I added -GroupMailbox to Identity and RemoveDelayHoldApplied based on your updates, but it might belong in more.

PR #3989, opened 2019-09-13
Update Move-CsUser.md

bjwhalen:
add content for UseOAuth switch which is for 2015 with CU8 and later.  Not needed for 2019.


MicrosoftHeidi:
This is confirmed for publish.

PR #4125, opened 2019-10-04
Additional updates for the new batch and group cmdlets that are in preview

tomkau:


tomkau:
Thanks @kenwith -- I stepped away and did not see the validation errors.  Otherwise would have fixed myself!

kenwith:
@tomkau - no problem, it is the schema that can be picky. :)

PR #4137, opened 2019-10-08
Fixed broken formatting for asterisk symbol

shivtorov:


kenwith:
@shivtorov - regarding this one, the same reference content shows up in terminal in PowerShell when you run  Get-Help <cmdlet>  ... so in those cases the HTML code you input won't render. I am trying to think how we could address it so that it works for both Markdown, HTML rendering on docs.microsoft.com, and text only rendering in PowerShell terminal.. what do you think?

shivtorov:
@kenwith, if `&ast;` is not suitable, then other options are:
1) escape asterisk with a backslash: `\*`, it will be \*
2) surround asterisk with spaces, so Markdown doesn't see it as a formatting character: `( * )`, it will be ( * )

kenwith:
@shivtorov - I like both ideas.. whatever you think is best let's go with that.. if you update your PR I will get it merged.

PR #4143, opened 2019-10-09
Highlight function on Start-SPOSiteRename.md

esjimenezp:
Is a good practie to highlight in the document that the feature is not enabled to all tenant, only for Preview

- [x] Update content


kenwith:
@esjimenezp - thanks for this call out to make it more prominent. One note is that the same reference content goes to   Get-Help cmdlet    so it will display in PowerShell terminal as text.. but I think it is fine to have the brackets around the word Note show.. just an FYI and something to keep in mind.

Issue #4131, opened 2019-10-07
Autocomplete has -DirectMembershipOnly


arowswell:
But this documentation doesn't list it an option?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 439f74a1-badd-e304-55d8-1981f86c5440
* Version Independent ID: e51876c4-c39d-a1b6-4624-599eb8f823d8
* Content: [New-DynamicDistributionGroup (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/users-and-groups/new-dynamicdistributiongroup?view=exchange-ps)
* Content Source: [exchange/exchange-ps/exchange/users-and-groups/New-DynamicDistributionGroup.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/users-and-groups/New-DynamicDistributionGroup.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws306:
Hi @arowswell, thank you for your feedback.

Could you please give me more detail about what you think is wrong with this article? Thanks!

arowswell:
If you load the module in PowerShell and type New-DynamicDistributionGroup -Direct then press <tab> to autocomplete, it shows -DirectMembershipOnly as a parameter. There is no mention of this on the docs page though.

dariomws306:
@arowswell thank you! Super clear now!

It looks like it's to avoid nested groups membership, but it doesn't work. I think it's still "under construction", I will add something to avoid confusion.

@chrisda maybe you can add something here, thanks!

yogkumgit:
@chrisda Please merge this PR #4163 , Thanks

yogkumgit:
@dariomws306  Thank you very much for the contribution and updating the documentation with PR. @arowswell Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #4163, opened 2019-10-11
Update New-DynamicDistributionGroup.md

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4131


chrisda:
Comments:

- When you manually add parameter descriptions, you must also include the parameter in the SYNTAX block (or blocks) at the top of the topic, and determining which parameter set (or sets) the parameter belongs in can be difficult without running [platyPS](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/repo_docs/UPDATE_CMDLETS.md). But, be advised that Get-Help and therefore platyPS hasn't worked in Exchange Online PowerShell for about 2 years now due to a Windows PowerShell bug.

- If a parameter is available in New-, then it's also very likely in Set-. The vast majority of cmdlets follow one of these two models.

  - Virtually every parameter that's available in New- is also available in Set-.

  or

  - The New- cmdlet is very bare compared to Set-, so most parameters are available on Set-.

  So, be very suspicious about new parameters that only exist in one or they other; they will likely be available in both.

- Use alpha sorting order for the parameters in the SYNTAX blocks and the location of the actual parameter description section. The only exception: required parameters are typically above optionals.

dariomws306:
Thank you! Should I update Set- or it's done?

Issue #4153, opened 2019-10-10
Looks like a generated Page, no value


FrankCarius:
Missing Description, missing parameter description. No statement, if it is also working with Teams Direct Routing. Should be updated 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a4d23360-4833-2386-b22d-4fe53448aa99
* Version Independent ID: 70dd7cb7-0eea-3adc-a962-a8f911a3522a
* Content: [Get-CsInboundBlockedNumberPattern (skype)](https://docs.microsoft.com/en-us/powershell/module/skype/get-csinboundblockednumberpattern?view=skype-ps)
* Content Source: [skype/skype-ps/skype/Get-CsInboundBlockedNumberPattern.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/skype/skype-ps/skype/Get-CsInboundBlockedNumberPattern.md)
* Service: **skypeforbusiness-powershell**
* GitHub Login: @tomkau
* Microsoft Alias: **tomkau**

yogkumgit:
@officedocsbot assign @yogkumgit

yogkumgit:
@kenwith Please helpus merge this PR #4169

yogkumgit:
@dariomws306  Thank you very much for the contribution and updating the documentation with PR. @FrankCarius Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

FrankCarius:
Only a few update

I think the „Applies to: Skype for Business Online“   should be “Applies to: Skype for Business Online, Microsoft Teams”

Frank



Frank Carius
Enterprise Architect / Partner

T:    +49 5251 304 600

Net at Work GmbH | Am Hoppenhof 32 A | 33104 Paderborn
Zentrale: +49 5251 304 600 | Fax: +49 5251 304 650
Handelsregister Paderborn: HRB 2663 | Geschäftsführer: Uwe Ulbrich
Näheres zur Informationspflicht nach Artikel 13 DS-GVO und zu unserer Verarbeitung Ihrer personenbezogenen Daten erfahren Sie unter https://www.netatwork.de/wp-content/uploads/Datenschutzinformation-für-Geschäftspartner_Net-at-Work_final.pdf
www.netatwork.de




Einführung von Office 365 geplant?
[...]

PR #4169, opened 2019-10-14
Patch 55

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4153

It should be reviewed. I think this information was removed some time ago, I don't know, maybe these cmdlets aren't fully supported.
There was a "Blocking incoming calls" article that was removed too.

Also, -CsTenantBlockedCallingNumbers are related and missing from docs so I think they should be added.


dariomws306:
> @dariomws306 - it could be that someone removed them if they aren't supposed to be used or are not functional. I have seen instances where parameters are in the code but not used. If a parameter is in the code, and shows up with [PlatyPS](https://github.com/PowerShell/platyPS) run, then I say we merge them and if they shouldn't be used then we can add a note saying as much. Going to go ahead and merge.

Thank you @kenwith! They exist in the code and seems to be working. I have only a doubt because I think these articles were completed some time ago, and also there was an article on the OfficeDocs-SkypeForBusiness repo with more detail about how to block number patterns. Maybe PG is deprecating them...

Issue #4270, opened 2019-10-29
OrganizationalUnit


adamfowlerit:
I think this needs to be clarified -for Exchange Online, you can't use an on-premises OU path even if the AD object lives in an OU. I can see people expecting this to work that way, e.g. https://www.reddit.com/r/PowerShell/comments/apfwu2/problem_with_getmailbox_organizationalunit/eg829s3/

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3b640892-b0a1-4f20-1e25-fe5f0e97f5b6
* Version Independent ID: 51774b36-90fa-1d9a-5800-1a54d9f70590
* Content: [Get-Mailbox (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/mailboxes/get-mailbox?view=exchange-ps#feedback)
* Content Source: [exchange/exchange-ps/exchange/mailboxes/Get-Mailbox.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/mailboxes/Get-Mailbox.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

yogkumgit:
@chrisda Please merge this PR #4286

yogkumgit:
@dariomws306  Thank you very much for the contribution and updating the documentation with PR. @adamfowlerit Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #4286, opened 2019-10-30
Patch 61

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4270

### DIFF
--- a/exchange/exchange-ps/exchange/mailboxes/Get-Mailbox.md
+++ b/exchange/exchange-ps/exchange/mailboxes/Get-Mailbox.md
@@ -591,6 +591,8 @@ Accept wildcard characters: False
 ```
 
 ### -OrganizationalUnit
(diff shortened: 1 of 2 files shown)


chrisda:
The OrganizationalUnit is 100% still available in Exchange Online. So, instead of acting like it's not, we need to come up with a description that solves the related issue (which I think is: in hybrid environments, you can't use the OrganizationalUnit parameter in ExO PowerShell to filter by OUs in on-prem AD).

The available OUs that you can use in Exchange Online PowerShell are extremely limited; literally, Get-OranizationalUnit only returns your own domain (e.g. contoso.onmicrosoft.com) and "Soft Deleted Objects" although I'm quite sure you cannot specify that as an OU to search (it doesn't error, but it returns no results, because you need to use the SoftDeletedMailbox switch instead).

PR #4296, opened 2019-11-01
Additional updates for batch and group policy assignment docs

tomkau:


tomkau:
Fixing issues then resubmitting

PR #4314, opened 2019-11-05
MD tweaks and a couple of tiny typos corrected.

greiginsydney:


kenwith:
@greiginsydney - just a note that HTML tags like ul and li won't render in the terminal with   Get-Help <cmdlet>   so people that use Get-Help from terminal will see the HTML tags.

greiginsydney:
Thanks @kenwith , I didn't realise.

Should I be zapping any HTML I encounter in commandlet documents in future (as well as of course not adding any of my own)?

kenwith:
@greiginsydney - probably a good idea but I would defer to you on how it would look. :) I will go ahead and merge this one to get your changes in and you can update if you want to remove them.

PR #4338, opened 2019-11-07
Update Get-CsVoicePolicy.md

amareshbAtMicrosoft:
We are removing this as per Roy (that owns voice scenarios). These should not be available for online.


amareshbAtMicrosoft:
Outlook<https://aka.ms/qtex0l> for iOS

PR #4341, opened 2019-11-08
adds note for beta on channel operations affected

rodiaza:
New functionality was added to support private channels, this is currently on /beta graph endpoint only.
We are adding a note for this if user is running the version on PS INT

### DIFF
--- a/teams/teams-ps/teams/Add-TeamChannelUser.md
+++ b/teams/teams-ps/teams/Add-TeamChannelUser.md
@@ -7,6 +7,10 @@ schema: 2.0.0
 
 # Add-TeamChannelUser
 
+> [!IMPORTANT]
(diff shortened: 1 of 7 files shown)


kenwith:
@rodiaza - the schema tests are super picky because the same content goes into the Get-Help from the terminal. So the content must match schema here: https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md   I am guessing the note needs to go in the Description area but I could be wrong. I have spent way to many hours trying to figure out cryptic error messages from schema checker. :)

bchopraMSFT:
@kenwith - I am good with the note added on the 7 powershell cmdlet docs.

rodiaza:
I ran the platyPS locally to find out which place would let it compile. I ended up moving the note to the description as @kenwith suggested

PR #4348, opened 2019-11-11
Clarify EXO V2 and usage with MFA

Batre-MSFT:
The note about MFA and EXO V2 is confusing and ambiguous. If someone is using MFA? Are there different steps to use new EXO V2? or They cannot use EXO V2? Please clarify


chrisda:
The added text is a question and doesn't belong in content. I'll follow up to get an answer.

PR #4367, opened 2019-11-13
Propose adding another Example

get-itips:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/3259


kenwith:
I know we were trying to be consistent on whether the "PS C:\>"  should be at the front of the command. My preference would be not to have it because if someone does a "Copy Code" from the web rendering then they have to backspace that out in order to run the command. However, I looked and the other examples have that so we should be consistent like you did.

Let's merge it and then circle back with the rest of the PowerShell team so we can get consistency regarding the "PS C:\>" prepended to beginning of examples across all PowerShell.

Issue #4368, opened 2019-11-13
Example two has an typo


sischanz:
When executing "Set-CsAuthConfig -Scenario BlockWindowsAuthExternallyAndModernAuthInternally" I receive this: 
Set-CsAuthConfig : Cannot bind parameter 'Scenario'. Cannot convert value "BlockWindowsAuthExternallyAndModernAuthInternally" to type

This works fine:
Set-CsAuthConfig -Scenario BlockWindowsAuthExternalyAndModernAuthInternally

Seems that Externaly is implemented with one l in that case

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 26656471-f4e4-3dbe-b915-5230400d7b7f
* Version Independent ID: 4dc006ee-6bf8-a929-94f7-499a5af1f7a1
* Content: [Set-CsAuthConfig (skype)](https://docs.microsoft.com/en-us/powershell/module/skype/set-csauthconfig?view=skype-ps)
* Content Source: [skype/skype-ps/skype/Set-CsAuthConfig.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/skype/skype-ps/skype/Set-CsAuthConfig.md)
* Service: **skypeforbusiness-powershell**
* GitHub Login: @hirenshah1
* Microsoft Alias: **hirshah**

yogkumgit:
@officedocsbot assign @yogkumgit

get-itips:
Hello @sischanz thanks for noticing.
We managed to report this to the product team and they are now aware of this, and we can expect some news from them too.
We do not tend (and like) to document a bug, but in this case, I was told to add a note until the cmdlet is fixed.
Also found this [support article](https://support.microsoft.com/en-us/help/4346673/new-cmdlets-to-manage-skype-for-business-server-2015-authentication) that documents the parameter with one L.
Thank you

PR #4397, opened 2019-11-18
Correcting parameter value

get-itips:
Please see https://github.com/MicrosoftDocs/office-docs-powershell/issues/4368

### DIFF
--- a/skype/skype-ps/skype/Set-CsAuthConfig.md
+++ b/skype/skype-ps/skype/Set-CsAuthConfig.md
@@ -49,10 +49,12 @@ This example forces all external clients to use MA but allows NTLM or Kerberos (
 
 ### Example 2
 ```powershell
-PS C:\> Set-CsAuthConfig -Scenario BlockWindowsAuthExternallyAndModernAuthInternally
+PS C:\> Set-CsAuthConfig -Scenario BlockWindowsAuthExternalyAndModernAuthInternally
 ```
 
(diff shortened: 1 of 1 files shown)


skype/skype-ps/skype/Set-CsAuthConfig.md
```diff
@@ -49,10 +49,12 @@ This example forces all external clients to use MA but allows NTLM or Kerberos (
...
 ### Example 2
 ```powershell
-PS C:\> Set-CsAuthConfig -Scenario BlockWindowsAuthExternallyAndModernAuthInternally
+PS C:\> Set-CsAuthConfig -Scenario BlockWindowsAuthExternalyAndModernAuthInternally
 ```
 
-This example blocks NTLM externally and MA internally. It forces all external clients to use MA, but allows all clients to use legacy authentication methods internally (even ADAL-capable clients).
+This example blocks NTLM externally and MA internally. It forces all external clients to use MA, but allows all clients to use legacy authentication methods internally (even ADAL-capable clients). 
+>[!NOTE]
+>Please note that in the current version of the module the parameter value BlockWindowsAuthExternalyAndModernAuthInternally is mispelled and to be able to use it, only one l for Externally should be used.
```
kenwith:
```suggestion
>Please note that in the current version of the module the parameter value BlockWindowsAuthExternalyAndModernAuthInternally is misspelled and to be able to use it, only one l for Externally should be used.
```

PR #4461, opened 2019-11-26

msbemba:
Added Record types and Added the As Job parameter.


chrisda:
@msbemba, I removed the AsJob parameter you added. Basically, every cmdlet has AsJob available to it in a remote PowerShell session (and all Exchange-based cmdlets operate in remote PowerShell). To have this parameter appear in some topics and not others causes more confusion.

PR #4479, opened 2019-11-28
Update exchange-online-powershell-v2.md

navgupta1208:


chrisda:
I moved the note about supported PS versions a bit lower in the topic. Technically, I would consider that new note as prerequisite, but I moved it just below the download link, where I still think it will be impactful.

Issue #4532, opened 2019-12-10
You can use this cmdlet to search message data for the last 10 days.


mrkeiths:
I believe the supported search timespan has been extended to 30 days.  There are a couple references in the "Description" section that mention 10 days.

FYI.  Get-MessageTraceDetail documentation also supports searching messages that were delivered within the last 30 days but no information to indicate acceptable timespan is included at all.

Thank you, Keith

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e3d7f9c1-78f0-5869-2e87-94c56c914316
* Version Independent ID: f7927615-71e8-069b-d2f3-0cae30092cd1
* Content: [Get-MessageTrace (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/mail-flow/get-messagetrace?view=exchange-ps)
* Content Source: [exchange/exchange-ps/exchange/mail-flow/Get-MessageTrace.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/mail-flow/Get-MessageTrace.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

yogkumgit:
@dariomws306  Thank you very much for the contribution and updating the documentation with PR. @mrkeiths Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

sysadmin-scs:
+1

can the article be updated?

chrisda:
The update was made, but it didn't take for some reason. I'm reapplying #5349.

sysadmin-scs:
Thanks for the quick response! There's one sentence that still needs updating:

"To search for message data that's between 10 and 90 days old, use the Start-HistoricalSearch and Get-HistoricalSearch cmdlets."

PR #4557, opened 2019-12-17
Patch 76

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4532


dariomws306:
@chrisda Could you please confirm that?

I'm trying in a customer tenant and I get this error:
"Invalid value for StartDate. StartDate cannot be more than 30 days from today."

But I can only return date from today...

PR #4561, opened 2019-12-18
Added 4 new cmdlets for number manipulation/translation rules

FilippSe:
Added 4 new cmdlets for number manipulation/translation rules to support translate caller and callee numbers for outbound and inbound calls to an alternate format functionality


FilippSe:
Need to fix merge issues

PR #4563, opened 2019-12-18
Fixed merge issues

FilippSe:


kenwith:
@FilippSe - metadata looks incorrect. Metadata is used when @tomkau creates the xml file that is checked into code so that `Get-Help` works in terminal. You can check with @get-itips for guidance. This is also written up here: https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/repo_docs/NEW_CMDLETS.md#parameter-metadata  You can check some of the other Skype cmdlets for reference to what Tom needs too.

PR #4589, opened 2019-12-30
Fix SipSignallingPort param name

jasomr:
GB to US English - removing one 'L' from the SipSignallingPort param name


get-itips:
Hello @jasomr 
How are you?

I checked and _SipSignallingPort_ is still a valid parameter/property

This fails:

image

This works
image

Am I missing something?

kenwith:
@jasomr - closing as it appears we aren't able to validate. Please reopen and let me know if we are missing something.

jasomr:
How it can be reopened? It was fixed in code just it takes ages the code reaches live …

@kenwith @get-itips

kenwith:
@jasomr - ok got you, it reopens automatically if you reply. Maybe we can just add a note @get-itips that the spelling changed? I will go ahead and merge now so the reference here reflects the new code.

get-itips:
> @jasomr - ok got you, it reopens automatically if you reply. Maybe we can just add a note @get-itips that the spelling changed? I will go ahead and merge now so the reference here reflects the new code.

Ok, I'll add a note 👍

get-itips:
Adding a note and referencing here the PR.

@jasomr Thank you for your contribution, please, take a look at [this](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/repo_docs/FAQ.md#i-am-a-microsoft-employee-but-new-to-github-what-should-i-do) so in the future we do not risk rejecting your PRs not knowing you are from Microsoft 👍

Issue #4672, opened 2020-01-13
Release notes are wrong


bagstac:
The Release notes state that version Version 0.3527.1 has been released.  It has been unlisted in the gallery and is no longer available. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: de2836a0-0af3-c147-2d4a-78220ac4e0f1
* Version Independent ID: 2cde9735-6281-bb9f-bf07-b30b2f6c7f37
* Content: [Exchange Online PowerShell V2](https://docs.microsoft.com/en-us/powershell/exchange/exchange-online/exchange-online-powershell-v2/exchange-online-powershell-v2?view=exchange-ps#feedback)
* Content Source: [exchange/docs-conceptual/exchange-online/exchange-online-powershell-v2/exchange-online-powershell-v2.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/docs-conceptual/exchange-online/exchange-online-powershell-v2/exchange-online-powershell-v2.md)
* Service: **exchange-online**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

yogkumgit:
@chrisda  Thank you very much for the contribution and updating the documentation with PR. @bagstac Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #4687, opened 2020-01-16
Update exchange-online-powershell-v2.md

navgupta1208:
Updating version number so that it is correct when the version is released.


chrisda:
Fixes Issue #4672

Issue #4708, opened 2020-01-20
The new Edge supports ClickOnce. 


ericlaw1979:
Use edge://flags to turn it on. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 65e527db-a4c2-782e-5a5f-ad8b3159a977
* Version Independent ID: f2de1272-5e37-c59f-fcb9-34fa66b0d268
* Content: [Connect to Exchange Online PowerShell using multi-factor authentication](https://docs.microsoft.com/en-us/powershell/exchange/exchange-online/connect-to-exchange-online-powershell/mfa-connect-to-exchange-online-powershell?view=exchange-ps#feedback)
* Content Source: [exchange/docs-conceptual/exchange-online/connect-to-exchange-online-powershell/mfa-connect-to-exchange-online-powershell.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/docs-conceptual/exchange-online/connect-to-exchange-online-powershell/mfa-connect-to-exchange-online-powershell.md)
* Service: **exchange-online**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws306:
Hi @ericlaw1979, thank you for your feedback.

I have tested it and works. I will propose an update to see if it will be approved.

yogkumgit:
@chrisda Please merge this PR #4714  Thanks

yogkumgit:
@dariomws306  Thank you very much for the contribution and updating the documentation with PR. @ericlaw1979 Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #4714, opened 2020-01-21
Update mfa-connect-to-exchange-online-powershell.md

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4708

I am proposing this update to see if you are agree as I'm not sure if ClickOnce is fully supported on Edge Chromium. As we need to enable it using edge://flags and there we can see this warning:
"WARNING: EXPERIMENTAL FEATURES AHEAD! By enabling these features, you could lose browser data or compromise your security or privacy. Enabled features apply to all users of this browser."


chrisda:
I simplified this greatly. I don't want this to turn into a laundry list of ClickOnce supported browsers (technically, Chrome supports ClickOnce with the right add-in). We can't just say IE, because IE is on the way out. Now that both new and old version of Edge support ClickeOnce, I changed it back to (for example, Internet Explorer or Edge). That's now accurate, and that's all we need to say.

Issue #4789, opened 2020-01-30
Command doesn't exist anymore


SanguptaMSFT:
This command doesn't exist anymore in the Skype for Business online module. please update.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: df5da894-a57f-d870-7a00-29507fb6d37f
* Version Independent ID: e52506bf-d874-625d-2b3e-c4d1c6f3ff54
* Content: [Get-CsTeamsMobilityPolicy (skype)](https://docs.microsoft.com/en-us/powershell/module/skype/Get-CsTeamsMobilityPolicy?view=skype-ps#feedback)
* Content Source: [skype/skype-ps/skype/Get-CsTeamsMobilityPolicy.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/skype/skype-ps/skype/Get-CsTeamsMobilityPolicy.md)
* Service: **skypeforbusiness-powershell**
* GitHub Login: @saragava
* Microsoft Alias: **saragava**

yogkumgit:
@officedocsbot assign @yogkumgit

get-itips:
Hello @SanguptaMSFT 
Thanks for noticing, we are adding a note thru a PR stating this.
Thank you

yogkumgit:
@get-itips  Thank you very much for the contribution and updating the documentation with PR. @SanguptaMSFT Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #4822, opened 2020-02-03
Adding removed note

get-itips:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4789


kenwith:
@get-itips warning is because the same content appears in terminal in Get-Help. Since it is only text in terminal it cannot render the note like docs.microsoft.com does. Maybe something like NOTE: so it stands out but there is no special text to render.. what do you think?

PR #4835, opened 2020-02-05
Removed Overridelobby setting documentation

nej-nej:
Removed the section containing the information about Allow Users to override lobby settings


LolaJacobsen:
@kenwith - @nej-nej is the PM for this area. Consider this TR'd.

Issue #4780, opened 2020-01-29
Does not completely support AD Synced Resources


auger282:
Set-Place only works completely with cloud created resources.

AD synced resources can only set: label, audiodevicename, videodevicename, displaydevicename, iswheelchairaccessible, floor, floorlabel, capacity, phone

These attributes are only available for cloud created resources: street, city, state, postalcode, countryorregion, geocoordinates, building

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws306:
Hi @auger282, thank you for your feedback.

I will test it next week and update the article accordingly.

dariomws306:
I have tested and you are right. If you try to modify these properties for synced rooms you will get "Encountered an internal server error" as them should be modified on-premises using AD or Exchange management tools.
The only property that I cannot say that you are right is Building, since I could modify it.

@chrisda I think the error message should be improved to something like "The operation cannot be performed because the object is being synchronized from your on-premises organization"

yogkumgit:
@chrisda Please merge this PR #4836 Thanks

auger282:
I see you've added the comments that it should be updated on-prem. These
options aren't available on-prem. Set-place is a cloud only command (as far
as I know).

dariomws306:
You shouldn't use Set-Place on-premises, use AD or Exchange management tools. But give me more time I'm doing more testing.

auger282:
Will do.. I appreciate the help. From what I've found the other attributes
do work in the cloud for set-place for the AD Synced accounts (odd but it
works). I wonder if this is a bug that they all don't work??

dariomws306:
Trying to get more information. I appreciate your patience.

auger282:
Any update? Is there a date in the future when I should check in?
Thanks for your help looking into this.

chrisda:
I updated the topic to say that the cmdlet only works on room mailboxes created in the cloud (not synchronized room mailboxes).

Thanks for the feedback.

Closing.

urbantom:
Hi, I think, the opinion "cmdlet only works on room mailboxes created in the cloud (not synchronized room mailboxes)" is incorrect. I have a cloud-only environment and it is affected by this issue.

dariomws:
Hi @urbantom, thank you for your comment.

Set-Place cmdlet article was updated and you can see this note now:
[...]

PR #4836, opened 2020-02-05
Update Set-Place.md

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4780


dariomws306:
@chrisda I think we should double check it, since I have just tried to modify these properties on-premises for a Remote Room Mailbox and when you run Get-Place you cannot see them with a value. So I think the information for this feature is stored in another place like custom resource properties on Exchange Server on-premises.

Maybe we should add something like "This property cannot be configured for rooms synced with Active Directory". What do you think? Or... Could you get more information?

Thanks!

chrisda:
Closing for further investigation. Well reopen as necessary

PR #4856, opened 2020-02-08
Corrected wrong example description

PowershellScripts:
The example 1 is about retrieving site design tasks, not scheduling them. 

The old description belongs to **Add**-SPOSiteDesignTask and not **Get**- and was probably a forgotten copy-paste.

### DIFF
--- a/sharepoint/sharepoint-ps/sharepoint-online/Get-SPOSiteDesignTask.md
+++ b/sharepoint/sharepoint-ps/sharepoint-online/Get-SPOSiteDesignTask.md
@@ -33,7 +33,7 @@ Used to retrieve a scheduled site design script. It takes the ID of the schedule
 
 ### Example 1
 
-This example applies a site design that includes two site scripts - totaling over 30 site script actions. Executing the commands will schedule the site design to be queued and run against the designated site collection.
+This example returns a scheduled site design whose ID is 501z8c32-4147-44d4-8607-26c2f67cae82 and which was applied on the site "https://contoso.sharepoint.com/sites/projectawesome".  
 
 ```powershell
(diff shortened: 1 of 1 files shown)


tseward:
I think scheduled is the wrong word for this. We don't 'schedule' site design applications, we run them and then can use this cmdlet to monitor status.

dariomws306:
> I think scheduled is the wrong word for this. We don't 'schedule' site design applications, we run them and then can use this cmdlet to monitor status.

I think this cmdlet is to get/return previously scheduled Site Design Task add/create with Add-SPOSiteDesignTask. The word scheduled is all around the article so I think it's ok.

From Add-SPOSiteDesignTask article:
"Similar to Invoke-SPOSiteDesign, this command is used to apply a published site design to a specified site collection target. It schedules the operation, allowing for the application of larger site scripts (Invoke-SPOSiteDesign is limited to 30 actions and subactions)."

tseward:
It's not a huge deal, but we don't "schedule" these tasks. We invoke them, which means start ASAP, but we can't say "run at 10 PM tonight".

Also, I see the explanation text is above the example, typically it is after the example.

get-itips:
@PowershellScripts Thanks for your contribution
Can you also change the title metadata of the md file? it says Add-SPOSiteDesignTask

I am agree that the word scheduled is not maybe 100% accurate however it was there all over the article...

I am agree that the description of the example should be after the example.

PowershellScripts:
About the word **schedule** - I agree with  @tseward that this is the behaviour we are seeing (speaking from user perspective), but that's the description used in every article on the topic, including PnP ones.  
  
On the description being above the example - it seems that a few articles follow this pattern. Would you like me to fix that?

kenwith:
@PowershellScripts - thank you for the work on this and the excellent collaboration. Sounds like this one is ready to go. Appreciate all the discussion to make the docs are the best they can be. I am going to use this as an example on why community driven, open source, content is the best option. :)

PR #4899, opened 2020-02-14
Document output of the function

holylander:
Basically adding some documentation about what the function returns.

### DIFF
--- a/sharepoint/sharepoint-ps/sharepoint-pnp/Get-PnPSiteCollectionAdmin.md
+++ b/sharepoint/sharepoint-ps/sharepoint-pnp/Get-PnPSiteCollectionAdmin.md
@@ -54,8 +54,10 @@ Required: False
 Position: Named
 Accept pipeline input: False
 ```
+
 ## Outputs
-**Microsoft.SharePoint.Client.User**
+
+### Microsoft.SharePoint.Client.User
 
 ## RELATED LINKS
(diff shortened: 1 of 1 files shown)


kenwith:
@holylander - looks like there is an error in the PR. I think it has to do with the ** surrounding but am not positive. You could try removing those and seeing if the schema checks pass.

On another note, the PnP cmdlet references are a little different than the rest of the Office cmdlet references. The PnP 'source' for the reference actually lives in the cmdlet code itself and each release the PnP maintainers generate the references from code and then copy the markdown over to this location so it is rendered alongside the rest of the Office cmdlet references. If we can get the schema checks to pass I will go ahead and merge this here but it will get overwritten the next time the PnP maintainers generate cmdlet reference from the cmdlet code unless it is also updated in the source for the cmdlet itself.

I believe the source file for this particular cmdlet is located here: https://github.com/SharePoint/PnP-PowerShell/blob/master/Commands/Site/GetSiteCollectionAdmin.cs

holylander:
Thanks a lot for the detailed explanation :) 

I have checked the link but I don’t see any markdown or documentation files on that repo, just standard code.

I will remove the bold md format and  see if the build success

holylander:
@kenwith I modified the commit so it uses `### `instead of `**` and seems like that did the trick

PR #4918, opened 2020-02-16
Update New-MoveRequest.md

vicktoria83:
In the code, I can see that we don't enforce the data loss threshold on migration service, so Accept... switch is not required anymore. In my tests it works like this. Hope, we can correct the article as it confuses customers. Thanks!


chrisda:
@vicktoria83, I believe you, but the AcceptLargeDataLoss parameter is available in 12 different cmdlets that span on-prem Exchange Server and Exchange Online. *Please reach out to me in email so we can discuss* (my GH alias is my internal alias). I'd like to keep the parameter descriptions consistent where they behave the same.

New/Set-MailboxExportRequest
New/Set-MailboxImportRequest
New/Set-MailboxRestoreRequest
New/Set-MoveRequest
New/Set-PublicFolderMigrationRequest
New/Set-PublicFolderMoveRequest

Closing for now.

Issue #4878, opened 2020-02-12
Clarification needed: Roles are not available in Security & Compliance


pkothree:
The roles Mailbox Search and Mailbox Import Export are only visible in the Exchange Admin Center (Permissions -> admin roles), but will not show up under the Security & Compliance portal.

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws306:
Hi @pkothree, thank you for your feedback.

Could you please kindly provide us with the article you are talking about?

pkothree:
[Search-Mailbox](https://docs.microsoft.com/en-us/powershell/module/exchange/mailboxes/search-mailbox?view=exchange-ps)

It should be made clear that this part references Exchange roles and are not visible in Security & Compliance.

> Note : By default, Search-Mailbox is available only in the Mailbox Search or Mailbox Import Export roles, and these roles aren't assigned to any role groups.

dariomws306:
There is a link on how to assign these roles and the instructions are for the Exchange Admin Center. Anyway, I arranged the format of the notes to make it more visible.

yogkumgit:
@chrisda Please merge this PR #4921 Thanks

chrisda:
@pkothree, I'm not sure I follow the connection between this cmdlet and the Security & Compliance Center. Nowhere in the topic does it state that this cmdlet is available in Security & Compliance Center PowerShell. It's only available in Exchange Server PowerShell and Exchange Online PowerShell.

Thanks for the feedback.

Closing

pkothree:
The problem I encountered was, that it is not stated that the roles are only available in Exchange, while a lot of other roles are available in the Security & Competence Center. I was asking if it would make more sense to clarify that a role, that is clearly important, is ONLY available in the Exchange Portal and would therefore deviate form a standard procedure for Office 365 admins.

chrisda:
@pkothree, yeah, the situation is not ideal. We're in this strange place where most of the Office 365 security features (anti-spam, anti-malware, anti-phishing, anti-spoofing, etc.) and some of the email related compliance features (like admin audit logging) are done in the Security & Compliance Center, yet the related cmdlets are and always have been in Exchange Online PowerShell (not Security & Compliance Center PowerShell).

[...]

PR #4921, opened 2020-02-17
Update Search-Mailbox.md

dariomws306:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4878


chrisda:
Sorry, I don't like using official Note call-outs in cmdlet reference topics because these same topics are also used for command line help. I also don't like using multiple Notes, Tips, Importants, whatevers back-to-back.

Closing

dariomws306:
Ok, maybe we should improve the format in another way, since it doesn't look good on docs.microsoft.com.

PR #4942, opened 2020-02-20
Align the help with current branding

skouti2:


chrisda:
Be advised that same change was also needed in the EnableRoomMailboxAccount parameter description on Set-Mailbox, which I did.

PR #5094, opened 2020-03-17
Update toc.yml

kaarins:
Fixed broken link in last pull request


kenwith:
@kaarins - I think you might be missing an href link after this line:  `- name: SharePoint in Microsoft 365` though I am not positive that is what is causing the warning.

PR #5101, opened 2020-03-18
Update Execute-AzureADLabelSync.md

samschan-ms:
Added description of cmdlet to sync labels to AAD

@SanjoyanM  @cabailey  to review


samschan-ms:
@chrisda FYI on the added description

cabailey:
Thank you, @samschan-ms! Quick question - wouldn't we normally say synchronize "from" and "to" (rather than "into Azure Active Directory")?  

I'm also wondering if we should include a reference/link to https://docs.microsoft.com/microsoft-365/compliance/sensitivity-labels-teams-groups-sites for context, but the downside is if we provide the title, it would need updating when the feature goes GA.  That's why I'm not a fan of putting the release status into the article title but I inherited this & seems to be common practice in my new team.

chrisda:
In general, I try to keep links in cmdlet reference topics to a minimum because these same topics are also used for help at the command line (`Get-Help <CmdletName>`), and the links aren't hot in the command line.
Judicious use of links is fine (too complex to explain in limited space, notices about features going away, etc.). If it's just FYI or adds no technical value, there's no need to add.
@samschan-ms, @cabailey,I'll leave it up to you to decide. You can always tweak the URL title in the link as if it were already GA.

PR #5148, opened 2020-03-24
Fix note box formatting.

MikeyMJCO:
The note box is incorrectly formatted along with a quote block.


chrisda:
@MikeyMJCO, there's a global problem with Notes that we're investigating, so I'm closing your PR.

Issue #5146, opened 2020-03-24
-WhatIf switch not correctly implemented


msbeane443:
The Remove-ApplicationAccessPolicy cmdlet removes the policy object when the parameter is specified.  Apologies if this is not the proper venue to report this behaviour.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9f40f0a0-5461-7821-c921-b0449ab229c1
* Version Independent ID: 5570e041-9b84-3c69-98e9-9ff414bb4a44
* Content: [Remove-ApplicationAccessPolicy (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/organization/remove-applicationaccesspolicy?view=exchange-ps#feedback)
* Content Source: [exchange/exchange-ps/exchange/organization/Remove-ApplicationAccessPolicy.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/organization/Remove-ApplicationAccessPolicy.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

dariomws:
Hi @msbeane443, thank you for your feedback. It's the right place to report documentation issues.

I'm updating these articles so that the reader knows that -Whatif is not available in these cmdlets.

yogkumgit:
@chrisda Please merge the PR #5166 Thanks

yogkumgit:
@dariomws  Thank you very much for the contribution and updating the documentation with PR. @msbeane443 Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5166, opened 2020-03-27
-Whatif parameter doesn't work on New/Set/Remove-ApplicationAccessPolicy.md

dariomws:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/5146


dariomws:
@chrisda Do you think we must report it? Thanks!

PR #5187, opened 2020-03-30
CELA updates for CsTeamsComplianceRecording cmdlets

aditdalvi:
Add a note from CELA that bots associated with the user's Teams recording policy are invited into the call or meeting to enforce compliance with the administrative set policy.

Remove the word "compliance" from any descriptions.


aditdalvi:
@kenwith can you add the necessary reviewers for this? Thanks!

PR #5202, opened 2020-04-01
Update Set-CsConferencingPolicy.md

MarcoCarampanta:
It's clearly outlined in the description:
"This parameter is used only in SfB Server"

But the "Applicable" field, shows SfB Online as well as SfB Server... which is incorrect and goes against the description. 

(Also confirmed that this is not available, on the SfB Online module)


kenwith:
@MarcoCarampanta - just wanted to first say thank you for taking the time to make the docs better and share your knowledge. Only a select few people take the time to do it and I really appreciate it.

@dariomws - I think this might also have to do with the transition to Teams but I cannot remember for certain. We can check with @tomkau if this is used in Teams or Skype for Business Online or if it is really only for Server.

dariomws:
@kenwith They should be available to users who are still using SfB Online until July 31, 2021. For Teams, there is another group of cmdlets: *-CsTeamsMeetingPolicy.

Issue #5207, opened 2020-04-01
-Archive parameter throws an error with Exchange Online


SeaGT:
Please provide an example of using -Archive to get info on an In-Place Archive mailbox.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 560849f0-488f-4ee7-2e91-64ab4598eb70
* Version Independent ID: 8fdde992-e4d8-1238-06ab-eb11857fddaa
* Content: [Export-MailboxDiagnosticLogs (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/mailboxes/export-mailboxdiagnosticlogs?view=exchange-ps#feedback)
* Content Source: [exchange/exchange-ps/exchange/mailboxes/Export-MailboxDiagnosticLogs.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/mailboxes/Export-MailboxDiagnosticLogs.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

yogkumgit:
@officedocsbot assign @yogkumgit

yogkumgit:
@chrisda Please merge this PR #5224 Thanks

yogkumgit:
@dariomws  Thank you very much for the contribution and updating the documentation with PR. @SeaGT Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5224, opened 2020-04-02
Archive parameter is available only in on-premises Exchange (Export-MailboxDiagnosticLogs)

dariomws:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/5207

image


dariomws:
There is something strange... It's listed under management role entries:
image
But if you return parameters from the code with Get-Command is not there:
image

Issue #5241, opened 2020-04-04
Get-CsOnlineDialInConferencingUserState not documented


shivtorov:
Empty help article

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 56234fd4-baa8-c376-7804-5e27e326a8c3
* Version Independent ID: 751d9ee5-ccee-bd18-4ce3-687a8c2769d2
* Content: [Get-CsOnlineDialInConferencingUserState (skype)](https://docs.microsoft.com/en-us/powershell/module/skype/get-csonlinedialinconferencinguserstate?view=skype-ps)
* Content Source: [skype/skype-ps/skype/Get-CsOnlineDialInConferencingUserState.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/skype/skype-ps/skype/Get-CsOnlineDialInConferencingUserState.md)
* Service: **skypeforbusiness-powershell**
* GitHub Login: @tomkau
* Microsoft Alias: **tomkau**

yogkumgit:
@officedocsbot assign @yogkumgit

get-itips:
Hello @shivtorov I am proposing thru a PR to add missing info. Thanks

yogkumgit:
@tomkau Please merge the PR #5266 Thanks

yogkumgit:
@islubin Please give your inputs in PR #5266 Thanks

yogkumgit:
@get-itips  Thank you very much for the contribution and updating the documentation with PR. @shivtorov Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5266, opened 2020-04-08
Adding missing info

get-itips:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/5241


shivtorov:
line 70 says "The user can be specified by using one of four formats: .... 3) the user's domain name and logon name, in the form domain\logon (for example, litwareinc\kenmyer); ...."

"domain\logon" format is not applicable here (at least in a sense meaningful for administrators)

PR #5278, opened 2020-04-13
modified permission needed to run the cmdlet

veronicageek:
Tested (_issue 2477 on the https://github.com/pnp/PnP-PowerShell/issues repo_) and unless this is going to change, then **Global Admin** rights are necessary to run this command.

### DIFF
--- a/sharepoint/sharepoint-ps/sharepoint-pnp/Get-PnPTenant.md
+++ b/sharepoint/sharepoint-ps/sharepoint-pnp/Get-PnPTenant.md
@@ -22,7 +22,7 @@ ResourceQuotaAllocated, and SiteCreationMode.
 
 Currently, there are no parameters for this cmdlet.
 
-You must be a SharePoint Online global administrator to run the cmdlet.
+You must be a Global administrator to run the cmdlet.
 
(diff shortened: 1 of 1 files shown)


kenwith:
@veronicageek - thank you for the PR. The PnP cmdlet references are a little different than the rest of the Office cmdlet references. The PnP 'source' for the reference actually lives in the cmdlet code itself and each release the PnP maintainers generate the references from code and then copy the markdown over to this location so it is rendered alongside the rest of the Office cmdlet references. I will go ahead and merge this here but it will get overwritten the next time the PnP maintainers generate cmdlet reference from the cmdlet code unless it is also updated in the source for the cmdlet itself.

I believe the source file for this particular cmdlet is located here: https://github.com/pnp/PnP-PowerShell/blob/master/Commands/Admin/GetTenant.cs

veronicageek:
Thank you @kenwith, I was a little bit troubled with the way it worked (2 different repos) so wasn't completely sure if that would be OK this way. I'm a contributor on the other repo so I'll ask the owner directly for the best way to do things in the future 👍🏻

PR #5279, opened 2020-04-13
Added note re: EOL'd OSes

mattpennathe3rd:
@chrisda Updating this at the request of David Maguire (as a result of a customer report, I believe).


mattpennathe3rd:
@chrisda Looks like a slight error was introduced during editing and propagated to other files. Second sentence reads, "To use this of Windows" (i.e., is missing the word "version").

PR #5335, opened 2020-04-20
Added parameter -PidfloSupported

FilippSe:
Added parameter -PidfloSupported


skype/skype-ps/skype/Set-CsOnlinePSTNGateway.md
```diff
@@ -372,6 +372,21 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -PidfloSupported
+Enables PIDFLO support on the PSTN Gateway. If turned on the .xml body payload is sent to the SBC with the location details of the user.
```
get-itips, line 376:
```suggestion
Enables PIDF-LO support on the PSTN Gateway. If turned on the .xml body payload is sent to the SBC with the location details of the user.
```

tseward: Will provide approval once this change is in place.

PR #5336, opened 2020-04-20
Added parameter -PidfloSupported

FilippSe:
Added parameter -PidfloSupported


skype/skype-ps/skype/New-CsOnlinePSTNGateway.md
```diff
@@ -403,6 +403,21 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -PidfloSupported
+Enables PIDFLO support on the PSTN Gateway. If turned on the .xml body payload is sent to the SBC with the location details of the user.
```
get-itips, line 407:
```suggestion
Enables PIDF-LO support on the PSTN Gateway. If turned on the .xml body payload is sent to the SBC with the location details of the user.
```

PR #5365, opened 2020-04-22
Update Get-MessageTrace.md

navesaims:
Get-MessageTrace only provides past 10 days results. Description field need to be modified to 10 days from 30 days.

## DESCRIPTION
You can use this cmdlet to search message data for the last 10 days. If you run this cmdlet without any parameters, only data from the last 48 hours is returned.

If you enter a time period that's older than 10 days, you won't receive an error, but the command will return no results. To search for message data that's between 10 and 90 days old, use the Start-HistoricalSearch and Get-HistoricalSearch cmdlets.


chrisda:
Already taken care of in #5366. Closing.

PR #5385, opened 2020-04-24
Add note that DisplayName filter is case-sensitive

skouti2:


kenwith:
@skouti2 - thank you for the PR and for taking the time to make the docs better. You sharing your experience and expertise is very much appreciated! :)

PR #5388, opened 2020-04-27
Update Set-CsTenantNetworkSite.md

carolynblanding:
@tomkau @chrisda @roykuntz 

Removing OVRP from Set-CsTenantSite as this is being hidden from PowerShell as it's not currently in use. 
LCSMain Change: 1970462

### DIFF
--- a/skype/skype-ps/skype/Set-CsTenantNetworkSite.md
+++ b/skype/skype-ps/skype/Set-CsTenantNetworkSite.md
@@ -49,10 +49,10 @@ The network region 'RegionRedmond' is created beforehand and 'MicrosoftSite1' wi
 
 ###-------------------------- Example 2 --------------------------
 ```powershell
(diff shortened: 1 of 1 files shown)


kenwith:
@get-itips - would you please confirm if it is removed from the latest module? @carolynblanding - if it is removed we can remove it here too, if it is still showing up in the code then we can add a qualifier that people should not use it even though they can see it in the module.

get-itips:
> @get-itips - would you please confirm if it is removed from the latest module? @carolynblanding - if it is removed we can remove it here too, if it is still showing up in the code then we can add a qualifier that people should not use it even though they can see it in the module.

I confirm that the parameter is still showing up for me.

kenwith:
@carolynblanding - I created a PR here: https://github.com/MicrosoftDocs/office-docs-powershell/pull/5393 and will add you as reviewer.

PR #5395, opened 2020-04-27
Updating timing of change to Teams client.

kenwith:


kenwith:
Delays with this process are also documented here: https://docs.microsoft.com/en-us/graph/teams-create-group-and-team

PR #5400, opened 2020-04-27
Update the New-CsTeamsComplianceRecordingApplication and Set-CsTeamsC…

ansrin:
…omplianceRecordinApplication cmdlet documentation

Update the New-CsTeamsComplianceRecordingApplication and Set-CsTeamsComplianceRecordinApplication cmdlet documentation - remove the warnings related to compliancerecordingpairedapplication


aditdalvi:
The diff is enormous!! Something went really wrong when you created the PR. Can you revert and do it directly from the github webpage instead of editing in notepad and uploading a new file (if that's what you did)?

ansrin:
Yeah, I followed steps as if we are to create new-cmdlet documentation but just didn't include the warnings. Maybe hence it shows so many diff changes, but really if you see there should be no new changes apart from removing those warnings.

Anand S

PR #5447, opened 2020-05-01
Exchange PowerShell flattening Step 3

dariomws:
Step 3: “Taking care of redirections”


chrisda:
Check the .openpublishing.redirection.json file; it uses a strange redirection syntax for the source path: 'exchange/virstualfolder/exchange' instead of the real 'exchange/exchange-ps/exchange' path. I updated the PR.

PR #5467, opened 2020-05-04
Update Get-PnPHubSiteChild.md

sympmarc:
Example 1 had misspelled cmdlet name: Get-PnPHubChildSite vs. Get-PnPHubSiteChild


kenwith:
Thanks Marc @sympmarc for your continued contributions! The PnP cmdlet references are a little different than the rest of the Office cmdlet references. The PnP 'source' for the reference actually lives in the cmdlet code itself and each release the PnP maintainers generate the references from code and then copy the markdown over to this location so it is rendered alongside the rest of the Office cmdlet references. I will go ahead and merge this here but it will get overwritten the next time the PnP maintainers generate cmdlet reference from the cmdlet code unless it is also updated in the source for the cmdlet itself.

I believe the source file for this particular cmdlet is located here: https://github.com/pnp/PnP-PowerShell/blob/master/Commands/Admin/GetHubSiteChild.cs

sympmarc:
I know that message! AFAIK, the examples aren't in the cs files?

kenwith:
I was pretty sure I had pasted in the response and I know you are an active contributor.. but then I thought I might as well paste it in for anyone that comes across this later in the future and reads it. :)  It is only the PnP that has those in the code.. the other cmdlet references have their one source of truth as here in GitHub. 

image

sympmarc:
Got it! I'll do a new PR with the fix.

PR #5473, opened 2020-05-04
Added notes

mukeshHCW:
These notes help users to know specific cases when to use which parameter

Proposed by Ian Liu

### DIFF
--- a/exchange/exchange-ps/exchange/federation-and-hybrid/Set-FederationTrust.md
+++ b/exchange/exchange-ps/exchange/federation-and-hybrid/Set-FederationTrust.md
@@ -82,7 +82,7 @@ Before you configure a federation trust to use the next certificate as the curre
 ### -Identity
 The Identity parameter specifies the name of the federation trust being modified.
 
-Note: For Office 365 or Exchange online tenants, use "Azure AD Authentication" as the input value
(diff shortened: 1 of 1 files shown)


mukeshHCW:
@kenwith I have been trying to add some notes here.. but it is giving me error.. how do i resolve this..

kenwith:
@mukeshHCW - it is the schema checker. It is something I have spent way to much time trying to figure out the errors too over the years. :)  The schema is so strict because the same markdown is converted to XML and checked into code so that Get-Help cmdlet works from the command line. The tool used to do that is an open source one called PlaytPS ( https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md ). Here is the schema it uses: https://github.com/PowerShell/platyPS/blob/master/platyPS.schema.md    I took a look and I think it has to do with the four pound signs.. I think it doesn't like those as headings because they don't match the schema.

mukeshHCW:
So if i remove those four pounds.. will it work?

kenwith:
It is worth a try. The schema validation error can be a pain to track down but the error in the build report can sometimes be helpful. 

"Failed to parse this MAML markdown file due to :84:(1054) '#### Note: For Office 365 or Exchange online tenan...' Expect Heading level 3, please use PlatyPS to do content validation"

Issue #5482, opened 2020-05-06
May cause confusion


hwkwicre:
"You must be a SharePoint Online global administrator to run the cmdlet."
Is that referring to a SharePoint administrator or a global administrator?
Thanks

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f494bf96-d8d8-4fff-b3dd-6a1ced1f2e1f
* Version Independent ID: eb9895b5-090d-7b07-ef3c-b90266aaf8e4
* Content: [Disconnect-SPOService (sharepoint-online)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-online/disconnect-sposervice?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-online/Disconnect-SPOService.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-online/Disconnect-SPOService.md)
* Service: **sharepoint-powershell**
* GitHub Login: @trent-green
* Microsoft Alias: **trgreen**

yogkumgit:
@officedocsbot assign @yogkumgit

get-itips:
Thanks for noticing, created a pr to correct this article and others. Thank you @hwkwicre

yogkumgit:
@get-itips  Thank you very much for the contribution and updating the documentation with PR. @hwkwicre Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5507, opened 2020-05-07
Removing global word from role

get-itips:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/5482

_SharePoint Online global administrator_

 role does not exists from available roles.


tseward:
@get-itips can we change this to 'SharePoint admin role' to match other documentation/the administration interface?

https://docs.microsoft.com/en-us/microsoft-365/admin/add-users/about-admin-roles?view=o365-worldwide

get-itips:
@tseward  I think we need to strip down the "role" word, if I use that sentence, it doesn't sound good:

_You must be a SharePoint admin role and a site collection administrator to run the cmdlet._

I could say _You must be assigned SharePoint admin role_  but that doesn't sounds good too.

I would propose using just _SharePoint admin_ ? as the UI and the role list says just that:

image

image

tseward:
@get-itips re-reading, it actually needs to be expanded upon.

"You must have the SharePoint Admin role or Global Administrator role [.....]"

PR #5545, opened 2020-05-12
Update Set-CsOnlineLisSwitch.md

danny-levin:


kenwith:
I think this was already merged in the previous PR. Let me know if that isn't correct and more updates still needed.

danny-levin:
Thx, it looks good.
(I changed a description of 2 parameters in 2 cmdlets (set switch & set port) and wanted both to include exactly the same description. Initially, it was not the case, hence I updated it again… - sorry about that…)

Issue #5609, opened 2020-05-19
link broken


hwkwicre:
Cmdlet Parameter Sets (https://go.microsoft.com/fwlink/?LinkID=187810).
Link doesn't work

For permissions and the most current information about Windows PowerShell for SharePoint Products, see the online documentation at https://go.microsoft.com/fwlink/p/?LinkId=251831 (https://go.microsoft.com/fwlink/p/?LinkId=251831).
Repeated link, second one doesn't work

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1ac87fde-0d91-d0cc-6c9e-444917652bec
* Version Independent ID: 10d9bf8c-a1ab-1cfe-c831-127d9d924fd2
* Content: [Remove-SPODataConnectionSetting (sharepoint-online)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-online/remove-spodataconnectionsetting?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-online/Remove-SPODataConnectionSetting.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-online/Remove-SPODataConnectionSetting.md)
* Service: **sharepoint-powershell**
* GitHub Login: @trent-green
* Microsoft Alias: **trgreen**

hwkwicre:
Related Links
Get-SPODataConnectionSetting
this one also doesn't work

yogkumgit:
@officedocsbot assign @yogkumgit

hwkwicre:
Related Links
Get-SPODataConnectionSetting
Also doesn't work

get-itips:
Thanks for reporting @hwkwicre

yogkumgit:
@get-itips  Thank you very much for the contribution and updating the documentation with PR. @hwkwicre Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5730, opened 2020-05-28
Fixing Several Links

get-itips:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/5609

### DIFF
--- a/sharepoint/sharepoint-ps/sharepoint-online/Remove-SPODataConnectionSetting.md
+++ b/sharepoint/sharepoint-ps/sharepoint-online/Remove-SPODataConnectionSetting.md
@@ -37,7 +37,7 @@ Remove-SPODataConnectionSetting -ServiceContext <SPServiceContextPipeBind> -Name
 
 This cmdlet contains more than one parameter set.
 You may only use parameters from one parameter set and you may not combine parameters from different parameter sets.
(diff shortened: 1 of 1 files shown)


get-itips:
One thing to note about this one

The cmdlets Get -SPODataConnectionSetting, Set -SPODataConnectionSetting, New -SPODataConnectionSetting and Remove -SPODataConnectionSetting **belong to SP Server but all but one, are in the sharepoint-online folder in the repo**

And also they have this notice inside of them:

>  This cmdlet applies to an on-premises environment only.
You cannot use this command in the SharePoint Online Management Shell.”

The related links fix should be considered a temporary fix until the root issue (cmdlets moved to correct folder) is fixed. Already reported to @kenwith 

Thanks

dariomws:
> One thing to note about this one
> 
> The cmdlets Get -SPODataConnectionSetting, Set -SPODataConnectionSetting, New -SPODataConnectionSetting and Remove -SPODataConnectionSetting **belong to SP Server but all but one, are in the sharepoint-online folder in the repo**
> 
> And also they have this notice inside of them:
> 
> > This cmdlet applies to an on-premises environment only.
> > You cannot use this command in the SharePoint Online Management Shell.”
> 
> The related links fix should be considered a temporary fix until the root issue (cmdlets moved to correct folder) is fixed. Already reported to @kenwith
> 
> Thanks

Nice, hope it´s fixed soon. All of them are in SPO with the exception of Get... Awesome!

Issue #5262, opened 2020-04-08
Please include the differences between this and the Teams Only tenants or at least link to the page.


Coldry:
This page states the following:
    Module: skype
    Applies to: Skype for Business Online

So where's the page that explains the details of this command for Teams? Ah... I hear you say, S4B online is Teams Telephony or is being migrated to Teams Telephone. Still, if the command works in a Teams only tenant then it should be included in the Applies to: pointing out the differences. 
OK having said that, I've noticed the "Get-Help Get-CsCallQueue -Full" (output below) command in a Teams Only tenant does return something different. The parameter type of the -Identity parameter shows as &lt;String&gt; not &lt;guid&gt;. Also, a number of the above parameters don't appear Deserialized in the Teams only tenant and there are a couple that isn't mentioned on this page which I guess the S4B Online doesn't have like "MusicOnHoldFileDownloadUri" or does it?

Name                                           MemberType   Definition
----                                           ----------   ----------
GetType                                        Method       type GetType()
ToString                                       Method       string ToString(), string ToString(string format, System.IFormatProvider forma...
DisplayAgents                                  NoteProperty Deserialized.System.Management.Automation.PSCustomObject DisplayAgents=
DisplayApplicationInstances                    NoteProperty string DisplayApplicationInstances=
DisplayDistributionLists                       NoteProperty Deserialized.System.Management.Automation.PSCustomObject DisplayDistributionLi...
DisplayStatistics                              NoteProperty string DisplayStatistics=Current queue size = 0
DisplayUsers                                   NoteProperty Deserialized.System.Management.Automation.PSCustomObject DisplayUsers=
OverflowActionTargetId                         NoteProperty Deserialized.System.Management.Automation.PSCustomObject OverflowActionTargetId=
PSComputerName                                 NoteProperty string PSComputerName=admingb1.online.lync.com
PSShowComputerName                             NoteProperty bool PSShowComputerName=False
RunspaceId                                     NoteProperty guid RunspaceId=
TimeoutActionTargetId                          NoteProperty Deserialized.System.Management.Automation.PSCustomObject TimeoutActionTargetId=
AgentAlertTime                                 Property     System.Int16 {get;set;}
Agents                                         Property     Deserialized.System.Collections.Generic.List`1[[Microsoft.Rtc.Management.Hoste...
AgentsCapped                                   Property     System.Boolean {get;set;}
AgentsInSyncWithDistributionLists              Property     System.Boolean {get;set;}
[...]

PR #5771, opened 2020-06-01
Update Set-CsExternalAccessPolicy.md

lindspea:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/4988
Added "applies to"


get-itips:
This is another occurrence of a long standing matter on cmdlets from SfB Online module that also apply to Microsoft Teams.
If we approve this change we would need to do the same for all of them.

Please see @kenwith comment [here](https://github.com/MicrosoftDocs/office-docs-powershell/issues/5262#issuecomment-623713910) as there is a current effort on trying to reorganize this content and recently it was decided to leave "applies to" as is.

dariomws:
Agree with @get-itips.

PR #5848, opened 2020-06-08
Update Set-AtpPolicyForO365.md

andypunt:
Adding EnableSafeLinksForWebAccessCompanion


dawander:
set-atppolicyforo365 -enablesafelinksforwebaccesscompanion $false or $true does not work (yet?).  We've tested across different tenants and regions with same result.  Error received is:  A parameter cannot be found that matches parameter name 'enablesafelinksforwebaccesscompanion'.
    + CategoryInfo          : InvalidArgument: (:) [Set-AtpPolicyForO365], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Set-AtpPolicyForO365
    + PSComputerName        : outlook.office365.com

Not opening a ticket with support but wanted to bring attention since docs says it is there.  FYI only.  Thanks.

Issue #5802, opened 2020-06-03
Missing module installation instructions


jreed5:
Only as FYI. Overview hyperlink is invalid or inaccessible publicly. The URL it points to is:
https://docs.microsoft.com/en-us/powershell/docs-conceptual/overview?view=whiteboard-ps

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 719f5227-2b47-3487-99bb-222d18d43552
* Version Independent ID: e7833ab4-d2b1-5840-4f2c-dda86599d3af
* Content: [Microsoft Whiteboard Admin](https://docs.microsoft.com/en-us/powershell/module/whiteboard/?view=whiteboard-ps)
* Content Source: [whiteboard/whiteboard-ps/whiteboard/whiteboard.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/whiteboard/whiteboard-ps/whiteboard/whiteboard.md)
* Service: **whiteboard-powershell**
* GitHub Login: @tbrosman
* Microsoft Alias: **tbrosman**

yogkumgit:
@officedocsbot assign @yogkumgit

yogkumgit:
@dariomws  Thank you very much for the contribution and updating the documentation with PR. @jreed5 Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5881, opened 2020-06-11
Fixed overview link in whiteboard.md

dariomws:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/5802

### DIFF
--- a/whiteboard/whiteboard-ps/whiteboard/whiteboard.md
+++ b/whiteboard/whiteboard-ps/whiteboard/whiteboard.md
@@ -8,7 +8,7 @@ title: Microsoft Whiteboard
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


kenwith:
I think @get-itips was fighting with these links too. 😁 I will merge and if it isn't right we can just keep trying. The links can be frustrating.

get-itips:
> I think @get-itips was fighting with these links too. 😁 I will merge and if it isn't right we can just keep trying. The links can be frustrating.

exactly :) I did this exact change some couple of days ago here https://github.com/MicrosoftDocs/office-docs-powershell/commit/9bce3aef1940fea936f00519b8d02db453a37efd#diff-affa0a8ee3ea001d863b32c12b18e2d5 but then was reverted here https://github.com/MicrosoftDocs/office-docs-powershell/commit/700a2f6ab953261ac2d79236a309d5bf7c194a80#diff-affa0a8ee3ea001d863b32c12b18e2d5 don't know exactly why.

Issue #5876, opened 2020-06-11
A few link issues


hwkwicre:
For more information, see Locale IDs Assigned by Microsoft (https://go.microsoft.com/fwlink/p/?LinkId=242911).
This link is broken

Please read (https://<kaarin to provide the link ) documentation to understand Information barriers with SharePoint Online.
https://<kaarin ??

For additional information about how to restrict a domain sharing, see Restricted Domains Sharing in Office 365 SharePoint Online and OneDrive for Business.
Is there supposed to be a link?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5d1c6b2b-5af6-acce-b7e8-bf784bb0973e
* Version Independent ID: 301586d3-ec04-c42d-4681-0bb59443d35c
* Content: [Set-SPOSite (sharepoint-online)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-online/set-sposite?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-online/Set-SPOSite.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-online/Set-SPOSite.md)
* Service: **sharepoint-powershell**
* GitHub Login: @trent-green
* Microsoft Alias: **trgreen**

yogkumgit:
@officedocsbot assign @yogkumgit

get-itips:
Thanks @hwkwicre

yogkumgit:
@get-itips  Thank you very much for the contribution and updating the documentation with PR. @hwkwicre Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5893, opened 2020-06-12
Fixing some links

get-itips:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/5876

@kaarins thanks for the links


get-itips:
according to Kaarin, links for information barriers will be published soon.

PR #5901, opened 2020-06-15
Update New-CsHostedVoicemailPolicy.md

hirenshah1:
-tenant attribute is not required for SFBO. SFBO is enabled for Cloud voicemail instead of Exchange UM

### DIFF
--- a/skype/skype-ps/skype/New-CsHostedVoicemailPolicy.md
+++ b/skype/skype-ps/skype/New-CsHostedVoicemailPolicy.md
@@ -23,7 +23,7 @@ This cmdlet was introduced in Lync Server 2010.
 
 ```
 New-CsHostedVoicemailPolicy [-Identity] <XdsIdentity> [-Description <String>] [-Destination <String>]
- [-Organization <String>] [-Force] [-InMemory] [-WhatIf] [-Confirm] [-Tenant <Guid>] [<CommonParameters>]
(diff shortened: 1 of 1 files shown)


get-itips:
The -Tenant parameter is not present in sfb 2019, if we agree in leaving the article as the last version it should be ok to remove it.

tseward:
If the parameter exists in previous versions of SfB Server, it should remain and the appropriate version(s) removed from the applicable to for the description switch.

dariomws:
I tried in 2015 with the latest updates and it´s there.

I also think that we should leave the parameter there, maybe remove 2019 from Applicable section, and think a note about EXO UM retirement could be a good idea.

skype/skype-ps/skype/New-CsHostedVoicemailPolicy.md
```diff
@@ -217,32 +201,6 @@ Accept pipeline input: False
...
-
-`Get-CsTenant | Select-Object DisplayName, TenantID`
-
-
-
-```yaml
-Type: Guid
-Parameter Sets: (All)
-Aliases: 
-Applicable: Lync Server 2013, Skype for Business Server 2015, Skype for Business Server 2019
```
tseward, line 237: Other changes need to be undone. Remove the appropriate Applicable version(s) from the `-Tenant` parameter.

yogkumgit:
@hirenshah1 Please implement the required changes and Merge. Thanks

yogkumgit:
@hirenshah1 Please implement the required changes and Merge. Thanks

PR #5910, opened 2020-06-16
Update New-CsBatchPolicyAssignmentOperation.md

raerupak:
$null is no longer an accepted value for PolicyName. To remove the currently assigned policy, use an empty string "".

### DIFF
--- a/teams/teams-ps/teams/New-CsBatchPolicyAssignmentOperation.md
+++ b/teams/teams-ps/teams/New-CsBatchPolicyAssignmentOperation.md
@@ -43,11 +43,11 @@ New-CsBatchPolicyAssignmentOperation -PolicyType TeamsMeetingPolicy -PolicyName 
 ```
 
 ### Example 2
-In this example, a policy is removed from a batch of users by passing $null as the policy name.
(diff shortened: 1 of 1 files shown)


teams/teams-ps/teams/New-CsBatchPolicyAssignmentOperation.md
```diff
@@ -43,11 +43,11 @@ New-CsBatchPolicyAssignmentOperation -PolicyType TeamsMeetingPolicy -PolicyName
 ```
 
 ### Example 2
-In this example, a policy is removed from a batch of users by passing $null as the policy name.
+In this example, a policy is removed from a batch of users by passing empty string "" as the policy name.
```
get-itips, line 46:
```suggestion
In this example, the current assigned policy is removed from a batch of users and the Global (Org-wide default) policy is applied by passing empty string "" as the policy name.
```

get-itips:
Suggesting to clarify which policy is applied when removing the current one.

tseward:
Also OK after the updates.

raerupak:
These changes are no longer needed.

PR #5929, opened 2020-06-17
Update Get-MessageTrace.md

v-padum:
The Get-MessageTrace cmdlet was recently changed to only support 10 days of data, as is specified in the "to search for message data that is greater than 10 days old" sentence - but the sentence mentioning data older than 10 but younger than 30 was not removed.

This change is to remove the confusion still left in the documentation.

### DIFF
--- a/exchange/exchange-ps/exchange/Get-MessageTrace.md
+++ b/exchange/exchange-ps/exchange/Get-MessageTrace.md
@@ -44,9 +44,7 @@ Get-MessageTrace
 ## DESCRIPTION
 You can use this cmdlet to search message data for the last 10 days. If you run this cmdlet without any parameters, only data from the last 48 hours is returned.
 
-If you enter a start date that is greater than 10 days and less than or equal to 30 days, only 10 days of data will be returned.
-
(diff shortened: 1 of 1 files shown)


chrisda:
@v-padum, I worked with a support engineer on this a couple of months ago. Can you confirm that the command will error if you enter a value greater than 10 days? That was the problem: it would happily take a value like 15 days without error, but the command would return no results.

v-padum:
Although you've merged this already, confirmed:
PS C:\Users\username> Get-MessageTrace -StartDate (get-date).adddays(-11) -EndDate (get-date)
Invalid StartDate value. The StartDate can't be older than 10 days from today.
    + CategoryInfo          : InvalidArgument: (:) [Get-MessageTrace], InvalidExpressionException
    + FullyQualifiedErrorId : [Server=YTOPR0101MB2155,RequestId=276f3e9e-599d-4dab-9106-6cf612585abb,TimeStamp=6/18/20
   20 2:24:03 PM] [FailureCategory=Cmdlet-InvalidExpressionException] A9FC7A86,Microsoft.Exchange.Management.FfoRepor
  ting.GetMessageTrace
    + PSComputerName        : outlook.office365.com

PR #5934, opened 2020-06-18
Adding two new Powershell Policies now available for Group Policy Ass…

cebulnes:
…ignment


dariomws:
Same as @get-itips (also tried with Get-CsGroupPolicyAssignment).

cebulnes:
> Currently on my demo tenant this cmdlet is giving me an access denied while trying to use it, with PS INT module version 1.0.25 @cebulnes Can you confirm the Teams tenant must also be enabled for some kind of private preview?

Yes, right now this is for whitelisted tenants, and will be in GA on June 30th

yogkumgit:
@chrisda Please merge this PR. Thanks.

PR #6032, opened 2020-06-27
Update Remove-TeamUser.md

markgort86:
The documentation is not clear enough, removing a member from the team does not remove the owner rights. You need to combine those commands to completely remove a user from a team. This PR is to update the documentation to make it more clear for people.

### DIFF
--- a/teams/teams-ps/teams/Remove-TeamUser.md
+++ b/teams/teams-ps/teams/Remove-TeamUser.md
@@ -39,7 +39,7 @@ In this example, the user "dmx" is removed the role Owner but stays as a team me
 ```
 Remove-TeamUser -GroupId 31f1ff6c-d48c-4f8a-b2e1-abca7fd399df -User dmx@example.com
 ```
-In this example, the user "dmx" is removed from the team.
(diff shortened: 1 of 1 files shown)


markgort86:
> I tested it using 1.0.7 module version and it removed the owner completely from the team. I think the example is correct.

Okay, the first thing i did wrong i was working on an older version of the module (my bad). With 1.0.7 it's not possible anymore to remove a member that is owner of the team. In version 1.0.3 where i was working on this command-let finished without any errors. Now i receive an error when i try the second example which make more sense. Can you tell me if there is something changed in the module about this behavior?

get-itips:
Yes, changes were made to this cmdlet between versions, however there isn't a detailed changelog.
I updated recently  this article with current behavior [here](https://github.com/MicrosoftDocs/office-docs-powershell/commit/abfde75fb78a4bd01f6e623d09e5546bbb324d58#diff-80e572404f9087461a8b9245f892b149)

markgort86:
@get-itips thanks for you're feedback on this PR this all makes more sense to me :-) With version 1.0.7 it's working as expected.

Issue #6054, opened 2020-07-01
Parameters


hwkwicre:
Some descriptions of parameters are quite confusing, such as
"Makes a SPO notification that the user is being move", "This parameter will force the cmdlet to execute only validation", "This parameter will only check for reserved users to be moved near to their sites"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a7dce14c-6e51-1144-4fd8-3add9af60e18
* Version Independent ID: b0a631e9-6afa-228d-1c4a-1912ee180604
* Content: [Start-SPOUserAndContentMove (sharepoint-online)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-online/start-spouserandcontentmove?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-online/Start-SPOUserAndContentMove.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-online/Start-SPOUserAndContentMove.md)
* Service: **sharepoint-powershell**
* GitHub Login: @trent-green
* Microsoft Alias: **trgreen**

yogkumgit:
@officedocsbot assign @yogkumgit

yogkumgit:
@trent-green Please help us in updating the PR #6070 . Thanks

PR #6070, opened 2020-07-02
Improving descriptions

get-itips:
@trent-green Trent, can you help me getting better descriptions for the following params?

-Reserved
-Notify

I appreciate.

closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/6054


get-itips:
@tseward Trevor, do you happen to know the description for Reserved and Notify on this cmdlet? I am almost certain that the current ones are wrong, I appreciate.

tseward:
@get-itips no however I will need to do this in the next couple of weeks so I may be able to get more info then.

get-itips:
> @get-itips no however I will need to do this in the next couple of weeks so I may be able to get more info then.

Ok great, let's merge this as it is waiting for a long time and later if we get the information for those params will make another pr.
@yogkumgit ready to be merged

PR #6139, opened 2020-07-10
Updates to case policy and retention rule cmdlets as per PM and github issues

markjjo:


chrisda:
@markjjo, I cancelled this PR and replaced it with #6141. As you can see from the build warnings, the fork or branch you were working out of was MEGA stale (it predated the Exchange cmdlet flattening back in early June and lots of fwlink removal and O365 to M365 updates that I did semi-recently).

PR #6204, opened 2020-07-23
Update New-Team.md

robinmalik:
Documentation references an old parameter name. Updated.

Also changed to reflect the actual default behaviour: Groups created using New-Team are created with HiddenFromExchangeClientsEnabled as $false. Previous documentation suggested the reverse behaviour. 

Added additional parameter which is often confused with the aforementioned one, and may be desirable: Hiding from the address list.


teams/teams-ps/teams/New-Team.md
```diff
@@ -12,8 +12,11 @@ ms.reviewer:
 
 ## SYNOPSIS
 This cmdlet lets you provision a new Team for use in Microsoft Teams and will create an O365 Unified Group to back the team. 
-Groups created through teams cmdlets, APIs, or clients will not show up in Outlook by default. 
-If you want these groups to appear in Outlook clients, you can use the [Set-UnifiedGroup](https://docs.microsoft.com/powershell/module/exchange/set-unifiedgroup) cmdlet in the Exchange Powershell Module to set the parameter HiddenFromExchangeClients to $false.
+Groups created through teams cmdlets, APIs, or clients will show up in Outlook by default.
```
dariomws, line 15:
```suggestion
Groups created through teams cmdlets, APIs, or clients will not show up in Outlook by default. 
```

teams/teams-ps/teams/New-Team.md
```diff
@@ -12,8 +12,11 @@ ms.reviewer:
 
 ## SYNOPSIS
 This cmdlet lets you provision a new Team for use in Microsoft Teams and will create an O365 Unified Group to back the team. 
-Groups created through teams cmdlets, APIs, or clients will not show up in Outlook by default. 
-If you want these groups to appear in Outlook clients, you can use the [Set-UnifiedGroup](https://docs.microsoft.com/powershell/module/exchange/set-unifiedgroup) cmdlet in the Exchange Powershell Module to set the parameter HiddenFromExchangeClients to $false.
+Groups created through teams cmdlets, APIs, or clients will show up in Outlook by default. 
+
+If you want to hide groups from Outlook clients, you can use the [Set-UnifiedGroup](https://docs.microsoft.com/powershell/module/exchange/set-unifiedgroup) cmdlet in the Exchange Powershell Module to set the parameter `HiddenFromExchangeClientsEnabled` to $true.
```
dariomws, line 17:
```suggestion
If you want these groups to appear in Outlook clients, you can use the [Set-UnifiedGroup](https://docs.microsoft.com/powershell/module/exchange/set-unifiedgroup) cmdlet in the Exchange Powershell Module to disable the switch parameter `HiddenFromExchangeClientsEnabled` (-HiddenFromExchangeClientsEnabled:$false).
```

teams/teams-ps/teams/New-Team.md
```diff
@@ -12,8 +12,11 @@ ms.reviewer:
...
+
+If you want to hide groups from the Global Address List (GAL), you can use the same `Set-UnifiedGroup` cmdlet to set the parameter `HiddenFromAddressListsEnabled` to $true.
```
dariomws, line 19:
```suggestion
```
Both (HiddenFromExchangeClientsEnabled and HiddenFromAddressListsEnabled) are true by default in Microsoft 365 groups, including when you run New-Team. I tested it on Teams PowerShell module version 1.1.4 (latest).

robinmalik: Thanks, I'll try that. I've been having to use 1.1.3 for my work because I require the ability to create "class" teams and I think it was failing under 1.1.4.

get-itips:
Agree w/ @dariomws

teams/teams-ps/teams/New-Team.md
dariomws: Thanks for collaborating on docs.microsoft.com. Let's also wait for my colleagues' feedback, I didn't test it on 1.1.3-preview, but I think we should have documented how it works on the latest stable version.

get-itips: Yes, Documentation always reflects stable version behavior.

teams/teams-ps/teams/New-Team.md
tseward: The default is that they do show up (i.e. if I provision a SharePoint Online Team site). Teams via the TAC (used to?) not show up.

yogkumgit:
@robinmalik Please update as per the feedback. Thanks

robinmalik:
@yogkumgit When I am next at a computer, I will do. This may be a few days. Please have patience.

teams/teams-ps/teams/New-Team.md
robinmalik: Having found time to test this today, `New-Team` is creating groups with both `HiddenFromAddressListsEnabled` and `HiddenFromExchangeClientsEnabled` set to true, and happy to report this applies to both module version 1.1.3 and 1.1.4. Odd that I was seeing different when I raised this but glad it's the default!

teams/teams-ps/teams/New-Team.md
robinmalik: Seems as though whatever the reason for me seeing groups with these settings set to false, that's not happening as of today (both are true, and thus hidden).

I'm finding the adjustment/suggestion flow here a bit difficult to read. Committing the two previous suggestions which seemed sensible, show as 'outdated'. I'll try committing this one @dariomws and then see if I can actually see what changes would be made to the file...

Issue #6258, opened 2020-07-31
Get-MessageTraceDetail documentation does not show required parameters or useful examples


MushyMiddle:
Get-MessageTraceDetail has required parameters, however the docs show all parameters as optional.  Minimally, it seems to require MessageTraceId, however it appears other fields are also required.

Further, the sole example isn't for Get-MessageTraceDetail, but rather for Get-MessageTrace.  While it does pipe the output to Get-MessageTraceDetail, that hardly helps explain how to use Get-MessageTraceDetail.

Finally, you should show an example of the output of the command - fields returned, explanations, etc.

Regarding output, a direct link to the returned type would help - why does every article force you to dig through the output types to find out what it returns?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6f784c39-5041-578f-7d9f-5b38310cc89c
* Version Independent ID: 4ab7a6db-db36-c736-6ce5-2f0e74b0cf6a
* Content: [Get-MessageTraceDetail (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/get-messagetracedetail?view=exchange-ps)
* Content Source: [exchange/exchange-ps/exchange/Get-MessageTraceDetail.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/Get-MessageTraceDetail.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

PR #6274, opened 2020-08-04
Updated mandatory parameters and added example to Get-MessageTraceDetail.md

dariomws:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/6258.


chrisda:
@dariomws, when you manually render a parameter from required to optional or vice-versa, be sure to update the syntax block (which you did) AND the Required: YAML tag value in the parameter description. Also, I like to move required parameters to the top of the syntax block, and the descriptions to the top.

Even so, doing this is dangerous. Required/Optional is determined by code (platyPS picks it up).

dariomws:
Thanks @chrisda. Sorry, I know it but missed it.

Move the required parameters to the top of the syntax block, and the descriptions to the top are new for me, but I think it's a good idea.

What happens here is that I think the parameters are not mandatory for PowerShell, but there is a strange code in the cmdlet to require them:
image

Issue #6283, opened 2020-08-06
TeamChannelUser CmdLets not available in 1.1.4


blindzero:
Hi,
starting MicrosoftTeams version 1.0.18 or better should support TeamChannelUser cmdlets.
I have 1.1.4 installed and CmdLets are not available.

```
ModuleType Name                            Version
---------- ----                            -------
    Script Connect-MS365                   1.3.0
  Manifest Microsoft.PowerShell.Management 3.1.0.0
  Manifest Microsoft.PowerShell.Utility    3.1.0.0
    Binary MicrosoftTeams                  1.1.4
    Script PackageManagement               1.4.6
    Script posh-git                        0.7.3
    Script PowerShellGet                   1.0.0.1
    Script profile                         0.0
    Script PSReadline                      2.0.0
```

```
CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Function        Get-CsBatchPolicyAssignmentOperation               1.1.4      MicrosoftTeams
Function        Get-CsGroupPolicyAssignment                        1.1.4      MicrosoftTeams
Function        Get-CsUserPolicyAssignment                         1.1.4      MicrosoftTeams
Function        New-CsBatchPolicyAssignmentOperation               1.1.4      MicrosoftTeams
Function        New-CsGroupPolicyAssignment                        1.1.4      MicrosoftTeams
Function        Remove-CsGroupPolicyAssignment                     1.1.4      MicrosoftTeams
Cmdlet          Add-TeamUser                                       1.1.4      MicrosoftTeams
Cmdlet          Connect-MicrosoftTeams                             1.1.4      MicrosoftTeams
Cmdlet          Disconnect-MicrosoftTeams                          1.1.4      MicrosoftTeams
Cmdlet          Get-CsPolicyPackage                                1.1.4      MicrosoftTeams
Cmdlet          Get-CsUserPolicyPackage                            1.1.4      MicrosoftTeams
Cmdlet          Get-CsUserPolicyPackageRecommendation              1.1.4      MicrosoftTeams
Cmdlet          Get-Team                                           1.1.4      MicrosoftTeams
Cmdlet          Get-TeamChannel                                    1.1.4      MicrosoftTeams
Cmdlet          Get-TeamsApp                                       1.1.4      MicrosoftTeams
Cmdlet          Get-TeamUser                                       1.1.4      MicrosoftTeams
[...]

PR #6301, opened 2020-08-10
Clarifying belongs to Preview Module

get-itips:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/6283

Replacing the note format as it is not rendered in get-help and also clarifying the state that this cmdlet belongs to the now called public preview module.

If this change is approved, will make the same change on the other cmdlets that also share this condition.


yogkumgit:

PR #6367, opened 2020-08-20

kme026:
Updating docs for new AllowRecordingStorageOutsideRegion Teams policy which is gonna be enabled on Monday.


get-itips:
Currently the link https://aka.ms/in-region does not work publicly, I do not know if it will be enabled soon.

Issue #6354, opened 2020-08-18
skype cmdlets missing Syntax section


jeffreyxrh:
There are various cmdlets in skype that are missing Syntax information, which will cause build errors.
Here is the build report:
![Capture](https://user-images.githubusercontent.com/22847442/90466599-c2c5f380-e144-11ea-8245-a4a1dc785c05.PNG)
The issue is not limited to only these files. Is this behavior expected?

get-itips:
Hello @jeffreyxrh 
I am completing the sintax of one of those articles so we can check if the build fails, can you share the full list ?
Thanks

jeffreyxrh:
There are 4 that I found:
skype\New-CsTeamsAppPermissionPolicy.md
skype\New-CsTeamsAppSetupPolicy.md
skype\Remove-CsTeamsPinnedApp.md
skype\Remove-CsVideoInteropServiceProvider.md

This one has an extra syntax tag: skype\Set-CsExternalUserCommunicationPolicy.md

get-itips:
@jeffreyxrh added commits to the PR

PR #6384, opened 2020-08-24
Adding syntax

get-itips:
closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/6354


get-itips:
> Maybe we should also complete the other sections.

These cmdlets have the following sentence "Intentionally not provided" and "The existence of this cmdlet is being documented for completeness, but do not use this cmdlet." so only adding the syntax so the build succeeds

yogkumgit:

PR #6390, opened 2020-08-25
[Hold] Inserted -name in Import-Module

NilsPeper:


yogkumgit:
@chrisda Please merge this PR. Thanks

chrisda:
@NilsPeper, may I ask why it's necessary to add Name? Name is positional, so you don't need to specify it, unless it fixes some issue or problem? I ask because if we're going to add it here, we're going to add it to the other two connection topics (ExO PowerShell and EOP PowerShell).

chrisda:
@NilsPeper, We've confirmed in #6565 that `-Name` is only required if you're trying to install and load the module in the same step. If you've already installed the module, you don't need to use `-Name` to load the module. I'll somehow convey this information in the related topics.

Thanks,

Closing

Issue #6356, opened 2020-08-18
Option : MigrateHiddenFiles is not available in preview (3.4.119.6)


takawazo:
In the public preview  (3.4.119.6),  MigrateHiddenFiles  is not available and if you specify this option, it will fail with error.
So, you must use IncludeHiddenFiles instead and confirmed the code and behavior.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 176873a5-a509-6337-0aa9-7db8ead2e863
* Version Independent ID: 73bb0f00-1909-6eb1-3a8c-247149480c9a
* Content: [Register-SPMTMigration (spmt)](https://docs.microsoft.com/en-us/powershell/module/spmt/register-spmtmigration?view=spmt-ps)
* Content Source: [spmt/spmt-ps/spmt/Register-SPMTMigration.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/spmt/spmt-ps/spmt/Register-SPMTMigration.md)
* Service: **spmt-powershell**
* GitHub Login: @serdarsoysal
* Microsoft Alias: **serdars**

PR #6394, opened 2020-08-25
Changing parameter name

get-itips:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/6356


spmt/spmt-ps/spmt/Register-SPMTMigration.md
```diff
@@ -270,7 +270,7 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
-### -MigrateHiddenFiles
+### -IncludeHiddenFiles
 If set to On, hidden system files will be migrated.
```
tseward, line 274: This is a bool. I'd recommend using true/false or $true/$false instead of On/Off.

yogkumgit:

PR #6413, opened 2020-08-27
Update Set-SPOTenant.md

robinmeure:
Modified the default value for the ShowEveryoneClaim as described here that the setting has changed since March 2018 https://docs.microsoft.com/en-us/office365/troubleshoot/access-management/grant-everyone-claim-to-external-users


yogkumgit:

Issue #6386, opened 2020-08-24
Need to document new Set-OrganizationConfig option OutlookGifPickerDisabled


philipzloh:
`OutlookGifPickerDisabled` is default `$false` and can be set to `$true`. This disables the GIF Search (powered by Bing) feature built into the Outlook Web Compose page.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 03895b30-2bb7-9e96-28db-def76ae59029
* Version Independent ID: 8056b3ee-4fab-3a5c-935b-a9d0e9ad82a4
* Content: [Set-OrganizationConfig (exchange)](https://docs.microsoft.com/en-us/powershell/module/exchange/set-organizationconfig?view=exchange-ps)
* Content Source: [exchange/exchange-ps/exchange/Set-OrganizationConfig.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/Set-OrganizationConfig.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

dariomws:
Hi @philipzloh, thank you for your feedback.

Are you from Microsoft?

PR #6434, opened 2020-08-28
Added OutlookGifPickerDisabled parameter to Set-OrganizationConfig.md

dariomws:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/6386.

@chrisda could you confirm this information? Maybe this parameter is not fully available yet because I cannot see the current value with Get-:
image


chrisda:
@dariomws, this feature must be in a controlled Preview, because I don't see it in an SDF tenant. I added a note about preview.

PR #6510, opened 2020-09-15
[Hold] Remove monikerRange info from files

byszhoubin:
We plan to enable v3 on this repo, however, the monikers info in yaml-header are not match the version info which configured in docfx.json, it will result in build failures.

This pull request is used to remove moniker info from markdown files to makre sure v3 build succeed.


chrisda:
@byszhoubin, the monikers in the topics are correct.  Per work items <https://ceapex.visualstudio.com/Engineering/_workitems/edit/231986> and <https://ceapex.visualstudio.com/Engineering/_workitems/edit/224630> the moniker values in the topics are correct. Therefore, instead of removing the monikers from all topics, I updated the docfx.json to match the moniker values in the topics.

Closing this PR.

mingwli:
@chrisda Thanks for the update, we'll take a try of the new config with docfx-v3.

qinezh:
@chrisda The PowerShell reference build pipeline doesn't support `monikerRange`, which is designed for other types of documentation. Can we clean up the `monikerRange` from the YAML header, which is blocking docfx v3 rollout?

PR #6534, opened 2020-09-17
Remove GrooveBlockOption from SyncClientRestriction

sakoo-ms:
Usage of GrooveBlockOption is discouraged.
Remove it from related SyncClientRestriction commands.


get-itips (changes requested):
I checked and `GrooveBlockOption` is still present in module version  16.0.20414.12000 in Set-SPOTenantSyncClientRestriction cmdlet, is it going to be removed soon? If not, the parameter needs to remain but a sentence added to it stating that the use is deprecated or discouraged.

The change against the PnP file must be done [here](https://github.com/pnp/PnP-PowerShell) or it will be overwritten (documentation for PnP is reflected from source code and synced later here)

Thank you

dariomws (commented):
Agree with @get-itips.

sakoo-ms:
@get-itips , it is planned to be deprecated. We would not want any new tenants use the option, so we would like to take them down from the document. We are also working on communicating to the business tenants using Groove regarding to the deprecation separately.

get-itips:
> @get-itips , it is planned to be deprecated. We would not want any new tenants use the option, so we would like to take them down from the document. We are also working on communicating to the business tenants using Groove regarding to the deprecation separately.
> 
> I will make a separate PR for PnP.

Ok then, if it will removed soon, i am ok with it. (otherwise people will create issues here stating that a parameter, present, is not documented.
Thank you

get-itips (approved):

sakoo-ms:
[PR on PnP project](https://github.com/pnp/PnP-PowerShell/pull/2929)

dariomws (approved):

sakoo-ms:
Decided to mention that it is deprecated instead of removing it. I will open another PR.

PR #6556, opened 2020-09-21
fix redirection rules

v-caxian:
- [x] [AB#305316](https://dev.azure.com/ceapex/Engineering/_workitems/edit/305316): yml redirections not supported in v2
  - `.yml` redirections with document_id are not supported in v2, they are exactly the same as setting to `false` .
  - In order not to change the generated document_id during migration, they need to be turned to `false`. 
  - However, the redirect_document_id: true will be supported after migrated to `docfx-v3`.


mingwli:
@dstrome  Could you please help review this PR? It's only about redirection change, and has non impact for the output.

PR #6939, opened 2020-12-03
Update Connect-ExchangeOnline.md

YuriySamorodov:


exchange/exchange-ps/exchange/Connect-ExchangeOnline.md
```diff
@@ -238,7 +238,7 @@ Accept wildcard characters: False
 ```
 
 ### -Prefix
-The Prefix parameter specifies an alias to add to nouns in the names of older remote PowerShell cmdlets (cmdlet with nouns that don't already start with EXO). A valid value is a text string without spaces, and you can't use the value EXO (this prefix is reserved for PowerShell V2 module cmdlets).
+The Prefix parameter specifies an alias to add to nouns in the names of older remote PowerShell cmdlets (cmdlet with nouns that don't already start with EXO). A valid value is a text string without spaces or special characters like underscrores, asterisks etc, and you can't use the value EXO (this prefix is reserved for PowerShell V2 module cmdlets).
```
YuriySamorodov, line 241: This is what has been changed

YuriySamorodov:
Errors

PR #6942, opened 2020-12-03
Update Grant-CsTeamsVideoInteropServicePolicy doc.

zihzhan:


chrisda:
@zihzhan, global build breaking problem fixed, but this specific topic is still having issues. I think it has to do with a) multiple code blocks in each ### Example section (there can be only one code block) and b) Text before the code block (not 100% sure about this one). I updated the topic accordingly.

zihzhan:
Thank you Chris for helping find and address those issues.
The document is in public and now looks really good to us!

PR #6948, opened 2020-12-05
Update Get-MailDetailATPReport.md

faoquong:
@chrisda


chrisda:
@faoquong, are we saying that the StartDate parameter only accepts a date value and not a date-time value? Your correction seems to imply that, but there are lots of other words in the description also need to be updated (which I can do no problem).

And is EndDate affected by this?

PR #6993, opened 2020-12-18
Update Initialize-PnPPowerShellAuthentication.md

LeonArmston:
Deleted -CertfificatePath from the documentation as this is not in the current PnP commandlet for Initialize-PnPPowerShellAuthentication and looks to be displayin the wrong data i.e. Password for the certificate being created.

I also think you should use -OutPath instead.


LeonArmston:
Incorrect

PR #7001, opened 2020-12-20
Update Set-AntiPhishPolicy.md

SignorelliDenis:
EnableAntispoofEnforcement parameter was renamed to EnableSpoofIntelligence according to Merged PR 1003345.
Related issue about this: https://github.com/MicrosoftDocs/microsoft-365-docs/issues/3713


chrisda:
@SignorelliDenis, approving, but always verify that the same parameter/condition doesn't also exist on the corresponding New- cmdlet. In this case this parameter is available on both New- and Set-.

chrisda:
I also made the same change in New-AntiPhishPolicy

PR #7079, opened 2021-01-14
Delete note

paul9955:
Per PM vikgup@microsoft.com, we need to delete the Note near the top of the page that states "This cmdlet is being replaced by the Set-MyAnalyticsFeatureConfig cmdlet." I've made the same change in the Get-UserBriefingConfig topic (https://docs.microsoft.com/powershell/module/exchange/get-userbriefingconfig?view=exchange-ps).


paul9955:
Thank you, Chris
- Paul

PR #7092, opened 2021-01-15
Update Get-SPOUser.md

pedossan:
Adding extra note related to the parameter -Limit which value "All" in which it may not output due to throttling

### DIFF
--- a/sharepoint/sharepoint-ps/sharepoint-online/Get-SPOUser.md
+++ b/sharepoint/sharepoint-ps/sharepoint-online/Get-SPOUser.md
@@ -95,7 +95,7 @@ Accept wildcard characters: False
 
 ### -Limit
 
-Specifies the maximum number of users returned. The default value is to return 500 users. To return all users specify the value "All".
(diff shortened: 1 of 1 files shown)


tseward:
@pedossan I'm not sure we should add this in here. It would require documenting other cmdlets and the behavior is highly scenario-specific (i.e. if you're performing a file migration to SPO at the same time you're attempting to run this cmdlet, there's a decent chance you'll see 429).

get-itips:
Also not sure about adding that note as @tseward says...

dariomws (commented):
I support the @tseward comment.

yogkumgit:
@pedossan

yogkumgit:
@pedossan Gentle reminder, please implement the feedback.

pedossan:
> @pedossan I'm not sure we should add this in here. It would require documenting other cmdlets and the behavior is highly scenario-specific (i.e. if you're performing a file migration to SPO at the same time you're attempting to run this cmdlet, there's a decent chance you'll see 429).

PR #7094, opened 2021-01-15
Update Get-SPOSite.md

WayneEwington:
Clarified that if the Template property is used in a filter, then this can be only the only condition with with either the -eq or -ne operator.


dariomws (commented):
It seems that Template works with more properties and also with -like/-notlike:
image

What do you think @get-itips @tseward?

get-itips:
> It seems that Template works with more properties and also with -like/-notlike:
> image
> 
> What do you think @get-itips @tseward?

Yes, I could use the `Template` parameter with `-notlike/-like` and also add another condition to the filter and it worked.
I know because I tested this with past versions that this wasn't working as is right now, so probably was fixed recently.

yogkumgit:
@WayneEwington

yogkumgit:
Gentle reminder, @WayneEwington please either reply to the feedback or close this PR.

WayneEwington:
Apologies with the delay. I have been reviewing this again the Engineering Manager (responsible for the cmdlet). The issue is that using -or in a filter results in a syntax error. Hence the documentation should reflect that.

I will close this and resubmit a different documentation change

Issue #7053, opened 2021-01-11
Code Example two has a missing Quotation Mark


who8e1:
The code Example 2 is missing a quotation mark at the end of the -ServerRelativeUrl Parameter 
`PS:>Move-PnPFile -ServerRelativeUrl "/sites/project/Shared Documents/Document.docx` **Here** `-TargetUrl "/sites/project/Archive/Document.docx" -OverwriteIfAlreadyExists`

This causes the code formatting to not display correctly
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cf06f8a6-0d6c-9c53-f015-57a5aebc59c9
* Version Independent ID: 7bc9ccc7-b9f9-bbae-7c80-aa5dedb0d3f2
* Content: [Move-PnPFile (SharePointPnPPowerShell)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-pnp/move-pnpfile?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-pnp/Move-PnPFile.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-pnp/Move-PnPFile.md)
* Service: **sharepoint-powershell**
* GitHub Login: @pamgreen-msft
* Microsoft Alias: **pamgreen**

get-itips:
Hello @who8e1 Ethan, How are you? hope you are doing great.

Even though the PnP documentation articles are on this repository, I need to kindly redirect you to open PnP cmdlets issues [here](https://github.com/PnP/PowerShell/), in that repository, where a member of the PnP community will be able to correct the article with your feedback.
The documentation for PnP cmdlets is then synced from that repo into this one.

Thank you!

yogkumgit:
@get-itips Thank you very much for the contribution and sharing this explanation. @who8e1 Hope this comment is helpful for you. If you see a documentation update is required, please feel free to open an issue for the same. We proceed here to close it. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #7101, opened 2021-01-17
Update Move-PnPFile.md

tseward:
Formatting fix/fix for #7053.


get-itips:
@tseward Trevor, AFAIK, if that didn't change recently, pnp documentation files are overwritten by the result of doc generation from PnP/PowerShell repo doc... so if we do this here it will be eventually overwritten.

tseward:
@get-itips ahhh that's right....

PR #7118, opened 2021-01-20
Update New-CsGroupPolicyAssignment.md

gosepulv:
OnlineVoiceRoutingPolicy is not currently supported via group policy, even though it appears to be. The service that consumes it is not group policy-aware.
Until that service is modified to support group policy assignment, the only workaround is to assign policies individually.
Whenever Teams service supports such assignment type, the article could be updated to include such PolicyType.


dariomws (approved):

tseward (approved):

yogkumgit:

PR #7136, opened 2021-01-21
Update Set-LabelPolicy.md to clarify advanced settings are for AIP client only

cabailey:


chrisda:
Also made corresponding change in New-LabelPolicy.

PR #7146, opened 2021-01-22
Update Set-M365DataAtRestEncryptionPolicy.md

tydakuja:


chrisda:
@tydakuja, be advised that platyPS picks up syntax blocks from code. I'm OK with manually removing it, but the better fix is to fix code. Not sure if/how RBAC applies to specific syntax blocks.

Issue #7113, opened 2021-01-20
In Step 5, unable to select registered app from Roles and Admin Assignments


singhkamall:
On [App Only Authentication](https://docs.microsoft.com/en-us/powershell/exchange/app-only-auth-powershell-v2?view=exchange-ps), I followed all 4 steps successfully however, I cannot add my newly created application as an assignment to **Roles and administrators** > **Exchange administrator** > **Add Assignment**.

My registered application:

![image](https://user-images.githubusercontent.com/31941421/105108327-3a42da80-5a77-11eb-81b3-a1460dc39360.png)

When I try to search it within **Roles and administrators** > **Exchange administrator** > **Add Assignment**:

![image](https://user-images.githubusercontent.com/31941421/105108548-bb9a6d00-5a77-11eb-8651-eea9bda86dfd.png)

singhkamall:
Looks like this doc is incomplete. I have solved my problem using this PowerShell command:

`Add-AzureADDirectoryRoleMember -ObjectId xxxxx -RefObjectId xxxxx`

After running it successfully, my app started appearing under Exchange Administrator assignments.

dariomws:
@singhkamall, this is how it looks now when you click on **Add assignments**. You need to click on **No member selected** and there you can select the app.
![image](https://user-images.githubusercontent.com/33589238/106122880-fcf5e100-6137-11eb-8be5-6ffdb1f87db1.png)
I'll update the article.

chrisda:
I just completed a major overhaul of [the topic](https://docs.microsoft.com/powershell/exchange/app-only-auth-powershell-v2) with lots of screenshots to match the current UI. I, to agree that several weeks ago when I started, something was amiss; Step 5 didn't work at all as described. Whatever mojo the Azure AD team was working through seems to have cleared up.

At least for me in my tenant. Please comment here if your UI doesn't match.

Also note, this issue is basically a duplicate of #7068.

Issue #7068, opened 2021-01-13
Step 5 UI changed. Can't find app in users list


yuri-wisestamp:
The UI looks different today:
![image](https://user-images.githubusercontent.com/5073294/104470151-10466f80-55c2-11eb-815c-f80549689bc4.png)

When clicking _here_, searching for Exchange Admin and clicking, we're getting to a screen similar to what the docs show:
![image](https://user-images.githubusercontent.com/5073294/104470474-6b786200-55c2-11eb-9f6b-bd739585064d.png)

 but cant find the registered app in the list, thus can't assign anything,

ZA-Ray:
I'm also battling to get this step working - glad I'm not the only one.  I was wondering if Azure AD Premium licensing is needed, but there is no mention of this on the page.

dg42xyz:
Also having this issue, any workaround available?

dariomws:
Hi @yuri-wisestamp, thank you for your feedback.

Could you kindly put the link to the article you are referring? I couldn't find it. Thanks!

ZA-Ray:
hi @dariomws , this is it:
[https://docs.microsoft.com/en-us/powershell/exchange/app-only-auth-powershell-v2?view=exchange-ps](url)

yuri-wisestamp:
@dariomws https://docs.microsoft.com/en-us/powershell/exchange/app-only-auth-powershell-v2?view=exchange-ps

dariomws:
@yuri-wisestamp @ZA-Ray @dg42xyz, this is how it looks now when you click on **Add assignments**. You need to click on **No member selected** and there you can select the app.
![image](https://user-images.githubusercontent.com/33589238/106122880-fcf5e100-6137-11eb-8be5-6ffdb1f87db1.png)
I'll update the article.

ZA-Ray:
Hi @dariomws, initially I thought I was using the wrong tool as I was in https://aad.portal.azure.com/ and your screenshot shows https://portal.azure.com/.  Navigating from https://aad.portal.azure.com/ I found i could search and add my app.  I then tried aad.portal.azure.com, and to my surprise again found that I could find my app.  This definitely wasn't working a few days ago when I first tried this.  ;-)  But happy it's working now!  Many thanks.

chrisda:
[...]

Issue #6865, opened 2020-11-17
Exchange isn't a selectable API permission


cclayton18:
In Step 2 we are told to select Exchange (shown in the screenshot under legacy APIs) and select Exchange.ManageAsApp, but no such card or permission is available to select.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d2cfafdf-cbe9-2ffe-eabe-becc66d51728
* Version Independent ID: 4a46c8a8-dc70-d877-271e-6679c465a6d5
* Content: [App-only authentication](https://docs.microsoft.com/en-us/powershell/exchange/app-only-auth-powershell-v2?view=exchange-ps#setup-app-only-authentication)
* Content Source: [exchange/docs-conceptual/app-only-auth-powershell-v2.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/docs-conceptual/app-only-auth-powershell-v2.md)
* Service: **exchange-online**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

chrisda:
Confirmed. Looking into it.

Smiliman:
@chrisda Thanks for your investigation - same confusion about missing Exchange.ManageAsApp here.

borismilincic:
I found the Exchange API when I changed to the tab "APIs my organization uses" and searched for the API ID (00000002-0000-0ff1-ce00-000000000000 in this case). This should be reflected in the article.

Other question: As this was listed in the "Supported legacy APIs" section before, is there a date when this will stop working? If so, where is it documented? Is there a way to use other APIs when using a service principal to manage Exchange Online?

ronnyholtmaat:
> I found the Exchange API when I changed to the tab "APIs my organization uses" and searched for the API ID (00000002-0000-0ff1-ce00-000000000000 in this case). This should be reflected in the article.
> 
> Other question: As this was listed in the "Supported legacy APIs" section before, is there a date when this will stop working? If so, where is it documented? Is there a way to use other APIs when using a service principal to manage Exchange Online?

That's probably only the case when you already are using the Legacy API. When you are **not using the Legacy Exchange API,** - in myt case - then it is **not** possible to select it. (also not under API's my organization uses).

Is there a Microsoft Graph API Alternative for App-only authentication for unattended scripts in the EXO V2 module?
[...]

PR #7176, opened 2021-01-28
Updated step 5 on app-only-auth-powershell-v2.md

dariomws:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7068.
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7113.


dariomws:
@chrisda Should I improve the size of the picture and add a border to it? Thanks!

chrisda:
@dariomws, I'm accepting this PR, but be advised that it does NOT completely fix the issues (which I thought were all being tracked under #6865).

Apparently, there are two different versions of Azure AD UI that you see depending on Azure AD Premium 1 vs. Azure AD Premium 2 licenses. The current content and your updates apparently only apply to AAD P1; In my UI (which is apparently AAD P2), starts out the same, but then falls apart:

1. In the Azure AD portal under **Manage Azure Active Directory**, click **View**.
2. Under **Manage**, select **Roles and administrators**.
3. Select one of the supported roles.

I get to this screen, when I can select a role, but there are absolutely zero assignment options to select. There's no flyout and the only thing I can do is view the text description of the selected role:

_NoWorkie

I'm working with an Azure PM who coincidentally reached out to me about this topic earlier this week.

Also note that I was in the process of updating this topic (including screenshots, which I haven't added yet) but I got to Step 5 and got stuck. That in-progress version is available here:

<https://github.com/MicrosoftDocs/office-docs-powershell/blob/AppAuthUpdates-chrisda/exchange/docs-conceptual/app-only-auth-powershell-v2.md>

As you'll see my initial thought was to leave 2 versions of the UI instructions in Step 5. This was before I knew about AAD P1 vs P2; I thought it was "old" and "new" UI. But if the AAD PowerShell instructions are available and identical in both cases, perhaps we should switch to those?

PR #7285, opened 2021-02-14
Update Set-SPOTenant.md

tseward:
Missing closing bracket for `[-ContentTypeSyncSiteTemplatesList MySites]`.


dariomws (commented):
@tseward I think they've done it in this way intentionally as the ExcludeSiteTemplate seems that should be used with ContentTypeSyncSiteTemplatesList. Anyway, I think we don't use to format it in this way. Maybe the ExcludeSiteTemplate should have their own description section.

get-itips:
Looks somewhat strange format so I ran a PlatyPS against the file, and I see that two parameters sets are being outputted

`### None (Default)`

and

`### ParameterSetContentTypeSyncSiteTemplatesList`

and that parameter is documented like this

`-ContentTypeSyncSiteTemplatesList <String[]> [-ExcludeSiteTemplate]`


Also, the parameter description section for ContentTypeSyncSiteTemplatesList **is not being rendered in the published version of the article**, because it lacks the yaml portion. Can you fix that?

get-itips (approved):

dariomws (approved):

yogkumgit:

PR #7324, opened 2021-02-19
Remove Exchange Online from EmailComposeMode list

philipzloh:
This is no longer supported in Outlook Web


chrisda:
@philipzloh, likely, but the parameter is still available in Exchange Online. So, until the parameter is actually removed from Exchange Online PowerShell, I want to nuance the availability message:

> This parameter is available or functional only in on-premises Exchange.

Issue #7314, opened 2021-02-18
what command for all mailboxes


WillemVerwijs:
a few examples were given. an example for all mailboxes would be great

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1c9a33ed-691e-83d5-41ed-606e3eb84e01
* Version Independent ID: f305f185-8ad2-1188-a448-5cfe52dadde6
* Content: [Get-MailboxFolderStatistics (ExchangePowerShell)](https://docs.microsoft.com/en-us/powershell/module/exchange/get-mailboxfolderstatistics?view=exchange-ps)
* Content Source: [exchange/exchange-ps/exchange/Get-MailboxFolderStatistics.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/Get-MailboxFolderStatistics.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

PR #7356, opened 2021-03-02
Added example to Get-MailboxFolderStatistics.md

dariomws:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7314.

### DIFF
--- a/exchange/exchange-ps/exchange/Get-MailboxFolderStatistics.md
+++ b/exchange/exchange-ps/exchange/Get-MailboxFolderStatistics.md
@@ -86,7 +86,8 @@ This example uses the IncludeAnalysis switch to view the statistics of Tony's Re
 
 ### Example 5
 ```powershell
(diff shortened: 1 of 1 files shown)


chrisda:
@dariomws, you'll notice that I altered the code significantly, but I tested it to confirm that it works.

Beware of `Get-Something |  Get/Set-SomethingElse` code syntax in Exchange PowerShell connections (which are always remote PowerShell). It certainly works, and conceptually, it's easy to understand, but I've seen it cause connection throttling even in surprisingly small organizations. It has to do with client-side vs. server-side filtering. Whenever I see that syntax, I always try to convert to:

```powershell
$Variable = Get-Something
$Variable | foreach {Get/Set-SomethingElse -IdentityOrWhatever $_.IdentityOrWhatever...}
```

It takes a bit of testing, but it typically works no problem.

Also, consider adding a similar example to Get-EXOMailboxFolderStatistics. But, what I previously explained is out the window for Get-EXO cmdets in the EXO V2 module; DON'T use the `$Variable = ...` syntax; use `Get-EXOMailbox <Parameters> | Get-EXOMailboxFolderStatistics <Parameters>.

dariomws:
Thank you @chrisda, I saw this using foreach, and now I know why. I'll add the example to Get-EXOMailboxFolderStatistics.

Issue #7363, opened 2021-03-03
PS module install command


nalhemyariirexorg:
[Enter feedback here]
is missing from the page, will be useful to add it so we can start using the commands

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6f98d029-8c45-af5e-1527-290690fa1283
* Version Independent ID: e7833ab4-d2b1-5840-4f2c-dda86599d3af
* Content: [MicrosoftWhiteboardAdmin Module](https://docs.microsoft.com/en-us/powershell/module/whiteboard/?view=whiteboard-ps)
* Content Source: [whiteboard/whiteboard-ps/whiteboard/whiteboard.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/whiteboard/whiteboard-ps/whiteboard/whiteboard.md)
* Service: **whiteboard-powershell**
* GitHub Login: @tbrosman
* Microsoft Alias: **tbrosman**

get-itips:
Hello, @nalhemyariirexorg thanks for reporting!

PR #7431, opened 2021-03-16
Adding installation instructions

get-itips:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7363


dariomws (changes requested):
Install command was there:
image

But I think the format is not nice. Maybe you can leave your new version and remove the old one.

get-itips:
@dariomws Good catch, wasn't expecting that, maybe because it is not enclosed in a code tag, will propose an improvement then

dariomws (approved):
Thank you!

tseward (approved):

yogkumgit:
@Kateyanne

Issue #7393, opened 2021-03-09
Connect-ExchangeOnline - ShowProgress:$false not behaving as intended


sanketdpm:
Hi,

As per the documentation of Connect-ExchangeOnline cmdlet:

> The ShowProgress parameter shows a visual progress bar in the PowerShell client module. The progress bar shows number of objects received and total number of objects requested. $false: The progress bar isn't displayed.

However, there is absolutely no difference between mentioning this parameter as false or skipping it (implicitly $true) Is this intentional or is there a disconnect between documentation and actual cmdlet?

Variations tried:
-ShowProgress:$false
-ShowProgress $false

cmdlet: Connect-ExchangeOnline
documentation: https://docs.microsoft.com/en-us/powershell/module/exchange/connect-exchangeonline?view=exchange-ps

Import-Module ExchangeOnlineManagement; Get-Module ExchangeOnlineManagement

ModuleType Version    PreRelease Name                                ExportedCommands
---------- -------    ---------- ----                                ----------------
Script     2.0.4                 ExchangeOnlineManagement            {Get-EXOCasMailbox, Get-EXOMailbox, Get-EXOMailboxFolderPermission, Get-EXOMailboxFolderStatistics…}

$PSVersionTable                                                                                                                                                                                                       

Name                           Value
----                           -----
PSVersion                      7.1.0
PSEdition                      Core
GitCommitId                    7.1.0
OS                             Darwin 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

chrisda:
Topic updated per #7466.

Thanks for the feedback.

Closing

PR #7466, opened 2021-03-19
Added note about ShowProgress in Connect-ExchangeOnline.md

dariomws:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7393.


dariomws:
@chrisda not sure if you agree with this issue. This is a well known issue, and I never saw this parameter providing any effect.

chrisda:
@dariomws, I'm cool with removing the use of the parameter from the example, because the progress bar for importing cmdlets is shown without it (it didn't a long time ago).

What I didn't realize was that ShowProgress $false doesn't hide the cmdlet import progress bar. This is discussed in the connection topic, so I'll remove it from there, too.

dariomws:
@chrisda thank you!

PR #7484, opened 2021-03-24
[Hold]Converted the arguments in the examples to use splatting.

Natfan:
[Splatting](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-7.1) is a method of passing a collection of parameter values to a command as a unit. Generally speaking, splatted parameters are easier to read and parse for humans.


chrisda:
@Natfan, great, but one of the main functions is to be able to copy/paste the example, which you lose when you go multi-line. Also, why stop here? There are likely hundreds and hundreds of examples in Exchange cmdlet reference topics that could use this same treatment.

I'm not convinced.

chrisda:
Closing

PR #7531, opened 2021-03-31
Update documentation for BusyOnBusyEnabledType parameter in Set-CsTeamsCallingPolicy cmdlet

isaiah-mayerchak:
Two new options (one supported: Unanswered, one currently unsupported: UserOverride) have been added as possible options for BusyOnBusyEnabledType. This PR updates the documentation for this parameter so that users can understand what the purposes/functions of these new options are.


JensTrier:
Any idea when this change will be live?

PR #7557, opened 2021-04-07

emmaryan123:
Updating the "Who can present" policy with the possible values, default values and type. Today the default value is "EveryoneInCompany" however, this is changing very soon to "Everyone" so i am updating to that. Thank you


emmaryan123:
@tomkau would you mind reviewing this and if all ok- merging? This is to update the possible values and default value. Today it is "EveryoneInCompany" but it is changing to "Everyone" so I'm updating the doc with this information. Thanks

emmaryan123:
@tomkau could you please review and merge? Thanks

PR #7587, opened 2021-04-12
Added note about -MediaRelayRoutingLocationOverride

FilippSe:
Added note about -MediaRelayRoutingLocationOverride

### DIFF
--- a/skype/skype-ps/skype/Set-CsOnlinePSTNGateway.md
+++ b/skype/skype-ps/skype/Set-CsOnlinePSTNGateway.md
@@ -15,6 +15,9 @@ ms.reviewer:
 ## SYNOPSIS
 Modifies the previously defined Session Border Controller (SBC) Configuration that describes the settings for the peer entity. This cmdlet was introduced with Microsoft Phone System Direct Routing.
 
+> [!NOTE]
+> -MediaRelayRoutingLocationOverride - this command is not available for use in direct routing scenarios.
+
 ## SYNTAX
 
 ```


dariomws (commented):
Based on your note, this parameter is not useful. This cmdlet is used for Direct Routing scenarios, so I think there are no more scenarios where the MediaRelayRoutingLocationOverride parameter can be used.

Maybe something like "this parameter is not available" or "This parameter is reserved for internal Microsoft use" as the parameter description instead of the note could be better.

dariomws (changes requested):
Some more history about this note, @FilippSe I think you took it from the following article/issue/PR:
https://github.com/MicrosoftDocs/OfficeDocs-SkypeForBusiness/issues/6823

I'm afraid I have to disagree with this sentence. Let's use something like what @TheBluLion suggested:
"This parameter is reserved for use with managed carriers."

For Set-CsOnlinePSTNGateway in the parameter description. For the other article as it's, in a note.

@lindspea could you kindly help here? Could you let us know why you added it in this way? Thanks!

@tseward @get-itips what do you think?

dariomws:
If you have a Microsoft alias, you can link your accounts in GitHub so that a blue "Microsoft submitter" label appears to any PR's or issue you submit. Guidance on it here: https://docs.opensource.microsoft.com/tools/github/index.html

Here is the site you can use to see the mappings between GitHub and Microsoft aliases: https://repos.opensource.microsoft.com/people?q=TheBluLion

FilippSe:
Hi,
I agree with having it marked as "This parameter is reserved for use with managed carriers."

Best regards,

Filipp Seljanko
Senior Program Manager, Microsoft
Email: ***@***.******@***.***>

lindspea:
@dariomws ok I see now. Let me know where I can help. Maybe the reviewers can change it as it's been merged already? Or let me know if must create a new PR from scratch.

dariomws:
@lindspea why you added this note instead of "This parameter is reserved for use with managed carriers."?

lindspea:
@dariomws  I'll look back on the issue to see what happened...Do you suggest removing the note, but adding the wording after this sentence " Note that Microsoft only recommends setting this parameter if the call logs clearly indicate that the default assignment of the datacenter for the media path does not use the path closest to the SBC datacenter. And then add: "MediaRelayRoutingLocationOverride is reserved for use with managed carriers" ?

dariomws:
@lindspea in the article you updated, I suggest replacing the following note:
> [!NOTE]
> -MediaRelayRoutingLocationOverride - this command is not available for use in direct routing scenarios.

With this one:
> [!NOTE]
> MediaRelayRoutingLocationOverride parameter is reserved for use with managed carriers.

@lindspea could you kindly confirm if you validated this incident/note with an SME?

@FilippSe could you kindly confirm that incident 226934920 really exist? I don't have permissions to check it. Thanks!

get-itips (changes requested):
Hello everyone.
Notes like that are usually added to the specific parameter description, also without a NOTE tag as that can't be rendered in `get-help` output.
Agree with @dariomws suggestion

dariomws:
@FilippSe do you think we are ok to go with:

"This parameter is reserved for use with managed carriers."

In the parameter description.

Thanks!

We also need to add this parameter to the New-CsOnlinePSTNGateway article because it seems to be missing.

FilippSe:
Hi,
Yes, I am good with this.

Best regards,

Filipp Seljanko
Senior Program Manager, Microsoft
Email: ***@***.******@***.***>

dariomws:
@FilippSe added a new PR to apply the update also to New-CsOnlinePSTNGateway. Please close this one. Thanks!

FilippSe:
Closed

Issue #7604, opened 2021-04-14
get-safelinksdetailreport case sensitivity


coproyer:
I've always been led to believe that PowerShell is case insensitive however, the documentation at https://docs.microsoft.com/en-us/powershell/module/exchange/get-safelinksdetailreport?view=exchange-ps does not state that case sensitivity for parameter values is a pre-requisite.  For example,

> `get-safelinksdetailreport -startdate 04-11-2021 -enddate 04-13-2021 -action blocked`

returns no results whereas

> `get-safelinksdetailreport -startdate 04-11-2021 -enddate 04-13-2021 -action Blocked`

returns expected results.

The documentation, including examples at the page and within `get-help get-safelinksdetailreport`, uses CamelCase however makes no notation that CamelCase is _required_.  Evaluation of either parameter values should be case insensitive or the documentation should be updated to reflect the CamelCase requirement.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 74c0d51f-873f-8a82-69b9-f27fb31b3eaf
* Version Independent ID: aad53fd4-dee4-bbb4-4135-2c974e688968
* Content: [Get-SafeLinksDetailReport (ExchangePowerShell)](https://docs.microsoft.com/en-us/powershell/module/exchange/get-safelinksdetailreport?view=exchange-ps)
* Content Source: [exchange/exchange-ps/exchange/Get-SafeLinksDetailReport.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/Get-SafeLinksDetailReport.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

chrisda:
@coproyer, that value should not be case sensitive. I suggest that you open a support case and see what they say. Report back here and we'll update the topic as necessary.

coproyer:
As expected, having access only to the O365 Support from the adminportal, that route returned totally anticipated results: "we'll notify the Exchange product team of the issue and in the meantime ask to have the documentation updated to note the case sensitivity issue."



Michael Royer
IT Security Officer
Department of Information Technology * City of Pasadena<http://www.cityofpasadena.net/>
+1 (626) 744-7929 * ***@***.******@***.***>

[...]

PR #7609, opened 2021-04-14
Update Get-SafeLinksDetailReport.md

sarah-nmachi:
Tried troubleshooting with a customer and figured he used a small letter B. Customer is Irate because he believes there should be a disclaimer as this is not the case for most exchange Online cmdlet


chrisda:
Fixed added text location that was causing build errors (valid location was within lines 80 to 90).

Related to Issue #7604

Issue #7636, opened 2021-04-18
Get-Team -DisplayName returns partial matches


rjohanneson:
Using the get-team -displayName "Team Name" returns teams that do not exactly match the supplied team name.

Example:
Get-Team -displayname "Art 4MS" -archived $false | select GroupID, DisplayName

GroupId                              DisplayName              
-------                              -----------              
deleted English Language Arts 4MS
deleted Art 4MS

Based on the documentation the command above should not have returned the first team listed in the results.

get-itips:
Hello @rjohanneson thanks for your feedback, I could reproduce so I will propose a different wording for the Get-Team article.
Thanks again

tos-kamiya:
Hi,

I have several teams that I manage with sequential numbers, and I am trying to add members with a script such as:

Connect-MicrosoftTeams
foreach($u in $users){
    Get-Team -DisplayName $team_Name | Add-TeamUser -User $u -Role Member
}

I didn't know that Get-Teams does fuzzy matching, so when I ran the above script, I got surprising results.

Can you please provide an option for exact matching other than this fuzzy matching?

tos-kamiya:
Solution?

I changed the team names from Software Engineering-1, Software Engineering-2, Software Engineering-3, to something like Software Engineering-Bunny, Software Engineering-Pony, Software Engineering-Penguin.

tos-kamiya:
Thank you for delving into the issue in such detail. While I personally have been mindful to avoid this behavior and haven't faced the same problem, I genuinely sympathize with others who have encountered the same issue.

As you rightly pointed out, it would be desirable to change this behavior. However, from what I've observed, it's often a traditional stance of MS not to alter behaviors, even if they are bugs, that might impact existing applications.

PR #7692, opened 2021-04-27
Clarifying Example 4 desc

get-itips:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7636

The search is not exact, so it is like you put "%Sales and Marketing%", I am open to other ways to describe this...

### DIFF
--- a/teams/teams-ps/teams/Get-Team.md
+++ b/teams/teams-ps/teams/Get-Team.md
@@ -59,7 +59,7 @@ Returns the team that matches the specified MailNickName
 ```
 PS> Get-Team -DisplayName "Sales and Marketing"
 ```
-Returns the team that matches the specified DisplayName
+Returns the team that includes the specified text in its DisplayName.
 
 ## PARAMETERS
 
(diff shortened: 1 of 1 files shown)


dariomws (approved):

tseward (approved):

yogkumgit:

PerePasqual:
The docs are misleading, and there are false statements.

First, it says:

"Returns the team that includes the specified text in its DisplayName. (This acts as a filter rather than an exact match)."

Then, if you go below, on the detailed info, it says this:

"Filters to return teams **with a full match to the provided displayname**. **As displayname is not unique, this acts as a filter rather than an exact match**. Note that this filter value is case-sensitive."

So 1) it's misleading that it says "with a full match to the provided displayname", and 2) in the second statement, "As displayname is not unique, this acts as a filter rather than an exact match", the first part of the phrase ("As displayname is not unique") can never be an explanation for this acting "as a filter rather than an exact match", because yeah, we know we can have several teams with the exact team name, and in that case it would make sense that
Get-Team -DisplayName would return all the Teams it found WITH THE EXACT MATCH; that's not a justification for returning partial matches, so a) you better remove that "explanation" and, b) and more important, remove the "WITH A FULL MATCH TO THE PROVIDED DISPLAYNAME" part.

But what really should be addressed is bug 7636 (https://github.com/MicrosoftDocs/office-docs-powershell/issues/7636), allowing to make both exact and partial searches.

PR #7717, opened 2021-05-05
Updated PollInterval parameter

Batre-MSFT:
PollInterval parameter is not applicable to Exchange 2013 and above, updated the Applicable list


chrisda:
@Batre-MSFT, whether it's supposed to be or not, or whether it works or not, the PollInterval parameter is available to me in TDS builds of E15 CU23, E16 CU20, and E19 CU9. I did a detailed search a couple of months ago to clean applicability up to match current reality.

Instead of removing the values from the Applicable metadata, I added the following text to the beginning of the description:

> This parameter is available or functional only in Exchange 2010.

This is a common situation, and the "available or functional" language is the best solution I've come up with for available parameters that aren't supposed to be available.

PR #7752, opened 2021-05-11

emmaryan123:
Updating documentation as "who can register" was added as an option


dariomws (approved):
WhoCanRegister also works with New-CsTeamsMeetingPolicy, and we have the following sentence in the description:
"This parameter is reserved for internal Microsoft use."

Maybe we should update it too.

PR #7805, opened 2021-05-20
Adding link and fixing two identical examples

brianreidc7:
Changed the two identical exampled to three more up to date ones and added a link to the replacement cmdlet as it was not there.


chrisda:
I removed the link. Links look terrible at the command line. Plus, if you're going to add a link in this cmdlet, we should also add them to every other *-ActiveSyncDevice cmdlet with that same note, but I don't want to do that.

PR #7831, opened 2021-05-26
Update New-OrganizationSegment.md

danmyhre:
Fixed syntax of example 3.


chrisda:
Same typo also existed in New-OrganizationSegment, so I also fixed it there in a different PR

Issue #7798, opened 2021-05-19
Does this only work for site collections, or can it target subsites too?


pyoungberg:
I successfully ran this command on my root site (https://contoso.sharepoint.com) ; however, I cannot seem to specify any of its subsites (https://contoso.sharepoint.com/hr ; https://contoso.sharepoint.com/ops ; https://contoso.sharepoint.com/field) and only news articles originating in the root site are getting the bolded tag (news posts from subsites of the root site are being ignored). Is this expected behavior? Does this command only work on site collections, or should I be able to run it on subsites as well?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1d50b175-d3d9-3eb2-68d3-e4cf84575487
* Version Independent ID: e03ea6c2-482f-8add-46dd-e83727af4fe5
* Content: [Set-SPOOrgNewsSite (SharePointOnlinePowerShell)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-online/set-spoorgnewssite?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-online/Set-SPOOrgNewsSite.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-online/Set-SPOOrgNewsSite.md)
* Service: **sharepoint-powershell**
* GitHub Login: @trent-green
* Microsoft Alias: **trgreen**

get-itips:
Hello @pyoungberg Paul, how are you?

Do you refer to this tag?
![image](https://user-images.githubusercontent.com/36666927/119521307-3df1b100-bd38-11eb-8ab1-62626d27eaff.png)

Tested this in my tenant, added a subsite with an URL like this: 

`set-SPOOrgNewsSite -OrgNewsSiteUrl https://m365x31200.sharepoint.com/sites/sc/TestC`

the confirmation message told me that the actual site added was the SC:

_The organization news site with url "https://m365x31200.sharepoint.com/sites/sc" is added to list._

Added a news post to "TestC" subsite and it appeared in the news, (tag wasn't present but the news post was there)

Please let me know if I understood you correctly,

pyoungberg:
@get-itips You're correct. Posts from the root site get tagged as indicated, but posts from its subsites don't, even when you run this command against each.

get-itips:
[...]

PR #7865, opened 2021-06-02
specifying

get-itips:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7798


tseward:
But we're not marking the "site collection" (which is just a container), but rather the root web. Honestly, I think "site" is OK.

get-itips:
> But we're not marking the "site collection" (which is just a container), but rather the root web. Honestly, I think "site" is OK.

To me, site is ok too, but agreed to create a PR with user proposal. I am ok if we leave as-is.

dariomws:
I think everyone agrees... Should it be closed?

Issue #7378, opened 2021-03-05
ConfigurationOnly backup often needs "-Force"


cloudgreen:
I believe some information about the following should be added:

If the farm is medium sized or bigger, you usually have to use the **-Force** parameter when doing a **ConfigurationOnly** backup. This is because SharePoint miscalculates the size of the backup to include basically everything, even though the **ConfigurationOnly** parameter is specified. This took me ages to figure out.

I am using SharePoint 2016.

For reference, see:
https://social.technet.microsoft.com/Forums/en-US/3356f9a3-8b5b-4970-b400-3da34fa69c4d/backup-sharepoint-2010-farm-configuration-settings-only-get-an-error-for-not-enough-disk-space-when?forum=sharepointadminprevious

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 59323fec-1d97-74d4-e815-ed2c941dd8bd
* Version Independent ID: c292dbf7-d1c5-7a3b-cc32-0d9b8fb06a8a
* Content: [Backup-SPFarm (SharePointServer)](https://docs.microsoft.com/en-us/powershell/module/sharepoint-server/backup-spfarm?view=sharepoint-ps)
* Content Source: [sharepoint/sharepoint-ps/sharepoint-server/Backup-SPFarm.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/sharepoint/sharepoint-ps/sharepoint-server/Backup-SPFarm.md)
* Product: **sharepoint-2016-powershell**
* GitHub Login: @Techwriter40
* Microsoft Alias: **pamgreen**

get-itips:
Hello @cloudgreen 
How are you? Actually, there is a note about that currently in the article:

> ### -Force
> Specifies to proceed with a backup operation even if the estimation is that the backup directory does not have sufficient space to hold the entire backup.

Thank you for your feedback.

cloudgreen:
Hi @get-itips!

All good, thank you. I hope you are well too. :)

The existing note is good, but it is too vague. I think it needs to be clearer that _-ConfigurationOnly_ often (always?) wrongly estimates the backup size to be way too big, and that the only way to be able to run the _-ConfigurationOnly_ backup at all is to also use _-Force_. This took me hours to figure out, and I'm not alone, as you can see by the forums... :)

I see two good options:

[...]

PR #7937, opened 2021-06-15
Clarifying

get-itips:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/7378


sharepoint/sharepoint-ps/sharepoint-server/Backup-SPFarm.md
```diff
@@ -200,6 +200,7 @@ Accept wildcard characters: False
 
 ### -ConfigurationOnly
 If set, the backup will contain configuration settings only (no data) for each of the objects to be backed up.
+In some cases that SharePoint estimates that it does not have sufficient space available, `-Force` parameter needs to be used in conjuntion to this parameter.
```
dariomws, line 203:
```suggestion
In some cases that SharePoint estimates that it does not have sufficient space available, `-Force` parameter needs to be used in conjunction with this parameter.
```

dariomws (commented):
The issue author wording looks good also. Something like:
"Note that this parameter often miscalculates the backup to be a lot bigger than it actually is, erroneously including content databases in the estimate, so -Force has then to be used in conjunction with this parameter."

Or there is some information you don't want to include?

tseward:
Sounds like this should be filed as a product bug instead, no? @TroyStarr thoughts?

yogkumgit:
@get-itips

get-itips:
> Sounds like this should be filed as a product bug instead, no? @TroyStarr thoughts?

Besides that (agreed) the sentence was already there in another part of the document.

TroyStarr:
Customers are welcome to open a support request through Microsoft Support to request an improvement to how the backup size estimation works.

That said, I think it's fine to provide additional guidance in the cmdlet help as well.  Would this wording be better as a separate paragraph in the description of the -ConfigurationOnly parameter?

SharePoint may overestimate the amount of space required for a configuration-only backup, which may result in SharePoint estimating that it doesn't have sufficient space to store the backup at the specified path. Use the `-Force` parameter in this situation to proceed with the backup if you have sufficient space to store the configuration-only backup.

get-itips:
Thanks @TroyStarr applied your suggestion

tseward (approved):

dariomws (approved):

yogkumgit:

PR #7941, opened 2021-06-16
Update Get-TenantAllowBlockListSpoofItems.md

fangsong21:


chrisda:
The content of these topics is also used at the command line for `Get-Help <cmdlet>`, so inline linking to other cmdlet topics isn't very helpful (all URLs are flattened into title-link). Also, I was told a long time ago not to use multiple levels of bullets (one level is fine).

PR #7942, opened 2021-06-16
Update New-TenantAllowBlockListSpoofItems.md

fangsong21:


chrisda:
For reference, the following YAML properties on any/all parameters don't populate correctly:

- Default value: None
- Accept pipeline input: False
- Accept wildcard characters: False

The values are ALWAYS None, False, False, even when they aren't. This has been an issue since the early Windows PowerShell days. No utility is able to read and pickup these values correctly. Any deviation from these values was added by hand to the topic by a human.

PR #7978, opened 2021-06-24
Remove -confirm parameter

michbrown-png:
-confirm is no longer a parameter then we need to update Start-CsExMeetingMigration

### DIFF
--- a/skype/skype-ps/skype/Set-CsCloudCallDataConnectorConfiguration.md
+++ b/skype/skype-ps/skype/Set-CsCloudCallDataConnectorConfiguration.md
@@ -68,21 +68,6 @@ The filtered collection is then piped to the `Set-CsCloudCallDataConnectorConfig
 
 ## PARAMETERS
 
-### -Confirm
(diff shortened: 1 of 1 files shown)


dariomws (changes requested):
@michbrown-png, please remove it from the syntax section.

Will it be removed in the next CU? I still see it on a server with the latest SU (May 2021 - 2046.369).

Start-CsExMeetingMigration doesn't have Confirm parameter on this server.

Thanks!

get-itips (changes requested):
agree with @dariomws suggestions

michbrown-png (commented):
Removed references to confirm in documentation

get-itips:
> Will it be removed in the next CU? I still see it on a server with the latest SU (May 2021 - 2046.369).

@michbrown-png Can you kindly answer @dariomws 's question here? Thank you!

michbrown-png:
@get-itips @dariomws @pupara It is already removed in the latest PowerShell Module release.

PR #8060, opened 2021-07-15
Add powerbimandatory as advanced setting for sensitivity labels

cabailey:


cabailey:
Helping out @paulinbar - establishing format to add other Power BI settings when released

PR #8152, opened 2021-08-06
Fix typo: extra bracket in syntax

Sophismata:
Removed an extraneous left bracket from one of the Get-Mailbox syntax blocks.


chrisda:
@Sophismata, believe it or not, it's not a typo. `[-Identity]` means the parameter is positional. Instead of running `Get-Mailbox -Identity chris` you can just do `Get-Mailbox chris`. `[-Parameter <Value>]` means the parameter is optional. So, when you put them both together, you get `[[-Identity] <MailboxIdParameter>]`.

Thanks for the feedback.

Closing

Sophismata:
Ah right — sorry I missed that. Thanks for the correction 😅.

PR #8173, opened 2021-08-12
Update Get-DynamicDistributionGroupMember.md

Benny-54:
This article is updated as per TASK 5331295. Thank you!

### DIFF
--- a/exchange/exchange-ps/exchange/Get-DynamicDistributionGroupMember.md
+++ b/exchange/exchange-ps/exchange/Get-DynamicDistributionGroupMember.md
@@ -123,10 +123,12 @@ This cmdlet supports the common parameters: -Debug, -ErrorAction, -ErrorVariable
 ## INPUTS
 
 ###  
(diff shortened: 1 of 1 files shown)


chrisda:
@Benny-54, that Inputs/Outputs topic hasn't been meaningfully updated in 3+ years. For new cmdlets, I never link to that topic, because the cmdlets aren't listed there. Only the old, venerable cmdlets are listed in that topic and are therefore linked to from the Inputs/Outputs sections in their cmdlet reference topics.

I cleaned all of this up ~3 years ago right after our migration from TechNet to docs.microsoft.com. Back then, the Inputs/Outputs sections of EVERY cmdlet reference topic linked to that topic. But, people complained if the Inputs/Outputs sections linked to that topic, but the cmdlet wasn't listed. I made sure that ONLY the cmdlets listed in that topic contained links to it in their Input/Output sections. There's also a separate version of this topic for Exchange 2010, and for the 2010 only cmdlets, I link to that topic.

But, I never add links in new cmdlet topics.

Closing

Benny-54:
@chrisda I didn't know this, thank you!

Issue #8011, opened 2021-07-01
How do we wait until the policy distribution is successful?


sguitardude:
The note in the third paragraph says, "wait until the policy distribution is successful."  How do we wait until the policy distribution is successful?  Are there commands we can run to see the current state of policy distribution?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: adaea894-fecc-16d3-0c4d-7971eda68b4b
* Version Independent ID: 4fe92cc7-84bf-2404-5481-4575b9d5d174
* Content: [Set-RetentionCompliancePolicy (ExchangePowerShell)](https://docs.microsoft.com/en-us/powershell/module/exchange/set-retentioncompliancepolicy?view=exchange-ps#feedback)
* Content Source: [exchange/exchange-ps/exchange/Set-RetentionCompliancePolicy.md](https://github.com/MicrosoftDocs/office-docs-powershell/blob/master/exchange/exchange-ps/exchange/Set-RetentionCompliancePolicy.md)
* Service: **exchange-powershell**
* GitHub Login: @chrisda
* Microsoft Alias: **chrisda**

dariomws:
Hi @sguitardude, thank you for your feedback and help us to improve docs.microsoft.com.

You can use Get-RetentionCompliancePolicy to verify the distribution status. For example, if you modify the following policy:
```Set-RetentionCompliancePolicy -Identity "Regulation 123 Compliance" -AddExchangeLocation "Kitty Petersen" -AddSharePointLocation "https://contoso.sharepoint.com/sites/teams/finance" -RemovePublicFolderLocation All -Comment "Added new counsel, 9/9/14"```
You can check the distribution status by running:
```Get-RetentionCompliancePolicy -Identity "Regulation 123 Compliance" -DistributionDetail | Format-List```

PR #8231, opened 2021-08-23
Added link to Set-RetentionCompliancePolicy.md

dariomws:
Closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/8011.


chrisda:
@dariomws, although it's tempting, do not shorten links in cmdlet reference topics. Although the shortened links will work on the web, when the topics are converted into XML and checked into code for help at the command line, the shortened links will not work; only the full and complete URL will work.

Because of this, you'll notice that there's no rule in the office-docs-powershell repo suggesting shortened links.

dariomws:
Thanks for the advice @chrisda!

PR #8363, opened 2021-09-20
Update TeamsCallParkPolicy

abnan:
Do not merge

### DIFF
--- a/skype/skype-ps/skype/New-CsTeamsCallParkPolicy.md
+++ b/skype/skype-ps/skype/New-CsTeamsCallParkPolicy.md
@@ -23,8 +23,7 @@ NOTE: The call park feature currently available in desktop. mobile and web clien
 
 ### Identity (Default)
 ```
-New-CsTeamsCallParkPolicy [-Tenant <System.Guid>] [-AllowCallPark <Boolean>] [[-Identity] <XdsIdentity>]
- [-Force] [-WhatIf] [-Confirm] [<CommonParameters>]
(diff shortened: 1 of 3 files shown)


dariomws:
Hi @abnan,

Is there still work in process? In that case, I think is better if we wait to review this PR. Thanks!

skype/skype-ps/skype/New-CsTeamsCallParkPolicy.md
```diff
@@ -102,6 +115,55 @@ Accept pipeline input: False
...
+
+Required: False
+Position: Named
+Default value: 99
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -ParkTimeoutSeconds
+Specify the number of seconds to wait before ringing the parker when the parked call hasn't been picked up. Value can be from 120 to 1800 (seconds).
```
dariomws, line 136:
```suggestion
Specify the number of seconds to wait before ringing the parker when the parked call hasn't been picked up. Value can be from 60 to 1800 (seconds).
```

image

abnan: Hi, the changes updating this value to 120 is currently rolling out and is available in lower rings.
image
Once it's available in all rings, I'll update this thread.

dariomws (approved):

dariomws:
@tiburd, I think this PR is ready to merge.

PR #8394, opened 2021-09-28
Add advanced settings for default sharing link to new/set-label

cabailey:


cabailey:
#hold-off until link is live

PR #8536, opened 2021-10-29
Update Set-Mailbox.md

bwinzenz:
We HAVE to stop referencing X500 addresses as X.500. There is no period between X and 500. There never has been one. X500 addresses are used to represent LegacyExchangeDN addresses across orgs. If someone enters a LegacyExchangeDN address as an X.500 address, it will NOT be recognized by Exchange!


chrisda:
@bwinzenz, I agree that the value should be X500, and it is. Your change is to the description of the value (basically, a comment). I was prepared to make sweeping changes everywhere for all X.500 references, but it appears that the actual name of the standard is X.500 (at least everywhere I can find it), so I don't see the harm in describing the X500 value as part of X.500 standard/directory services.

Or am I missing something?

chrisda:
@bwinzenz, per our earlier IM conversation, I'm closing this PR.

PR #8561, opened 2021-11-05
Update Get-CsHuntGroup.md

CLYVR:
Adding caution at top of page indicating that this cmdlet has been deprecated


CLYVR:
@CarolynRowe Please review.  Updating old cmdlets to add a caution at top of each page to say cmdlet is deprecated and pointing to the new cmdlet.  If there are any issues please let me know.  Thanks!

PR #8562, opened 2021-11-05
Update Get-CsOnlineUser.md

isrumnon:
Updating documentation to reflect changes in TPM 3.0 onwards


dariomws (changes requested):
Same here, please review Sean comment.

And I think you should add the attributes deprecation note to the description section.

PR #8564, opened 2021-11-05
Update Get-CsOnlineVoiceUser.md

isrumnon:


skype/skype-ps/skype/Get-CsOnlineVoiceUser.md
```diff
@@ -103,6 +103,15 @@ Accept pipeline input: False
...
 ```
 
+```yaml
+The below updates are applicable to TeamsOnly tenants using TPM modules 3.0 onwards. Attributes are deprecated because they are no longer relevant to TeamsOnly tenants:
+- Force
+- NumberOfResultsToSkip
+- CorrelationId
+- Verb
+- ResultSize
+```
```
sdwheeler, line 113: Yaml code blocks in the reference are only used for parameter metadata. Follow the pattern for the existing content. What you have here is straight up markdown. It does not belong in a Yaml code block. The description for the parameter goes before the Yaml metadata block.

Also, the way that this is described, it is unclear whether the list of attributes is the allowed list or the deprecated list.

skype/skype-ps/skype/Get-CsOnlineVoiceUser.md
```diff
@@ -193,6 +202,11 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+```yaml
+The below updates are applicable to TeamsOnly tenants using TPM modules 3.0 onwards. 
+LiscenceState has been deprecated because it is no longer relevant to TeamsOnly tenants
+```
+
```
sdwheeler, line 209: Same comments apply here. 

What does **LicenseState** have to do with the **Identity** parameter? This seems out of place.

sdwheeler (changes requested):
See my comments below.

PR #8601, opened 2021-11-15
Removed ExternalInOutlookEnabled and added link to Set-OrganizationConfig.md

dariomws:
Closes https://github.com/MicrosoftDocs/OfficeDocs-Exchange/issues/2860.

### DIFF
--- a/exchange/exchange-ps/exchange/Set-ExternalInOutlook.md
+++ b/exchange/exchange-ps/exchange/Set-ExternalInOutlook.md
@@ -136,3 +136,5 @@ This cmdlet supports the common parameters: -Debug, -ErrorAction, -ErrorVariable
 ## NOTES
 
 ## RELATED LINKS
+
+[Set-OrganizationConfig](Set-OrganizationConfig.md)
--- a/exchange/exchange-ps/exchange/Set-ExternalInOutlook.md
+++ b/exchange/exchange-ps/exchange/Set-ExternalInOutlook.md
@@ -137,4 +137,4 @@ This cmdlet supports the common parameters: -Debug, -ErrorAction, -ErrorVariable
 
 ## RELATED LINKS
 
-[Set-OrganizationConfig](Set-OrganizationConfig.md)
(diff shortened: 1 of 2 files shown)


chrisda:
@dariomws, see my comments in the related Issue (I don't believe Set-ExternalInOutlook and MailTipsExternalRecipientsTipsEnabled on Set-OrganizationConfig have anything to do with each other).

The removal of `[-ExternalInOutlook]` from the syntax block is fine.

Two additional things to remember:

- Availability of a cmdlet/parameter is different from functionality. The applicable/Applicable metadata values on cmdlets/parameters needs to accurately reflect if the cmdlet/parameter is AVAILABLE in a particular environment. We can use additional text (the parameter description, the SYSOPSIS section, and/or the DESCRIPTION section) to explain that the cmdlet/parameter doesn't work in a specific environment. In other words: if a cmdlet/parameter is AVAILABLE, don't remove it.
- Always use full docs.microsoft.com topic URLs in cmdlet reference topics. Relative topic links will work fine on the web, but they won't populate correctly when the topics are converted from markdown to MAML XML for code check-in and use at the command line.

dariomws:
Thank you @chrisda. It was my mistake, I didn't see that one parameter is for inbound and the other one for outbound.

> * Availability of a cmdlet/parameter is different from functionality. The applicable/Applicable metadata values on cmdlets/parameters need to accurately reflect if the cmdlet/parameter is AVAILABLE in a particular environment. We can use additional text (the parameter description, the SYSOPSIS section, and/or the DESCRIPTION section) to explain that the cmdlet/parameter doesn't work in a specific environment. In other words: if a cmdlet/parameter is AVAILABLE, don't remove it (including its applicable/Applicable value).

I remember this one, thank you.

> * Always use full docs.microsoft.com topic URLs in cmdlet reference topics. Relative topic links will work fine on the web, but they won't populate correctly when the topics are converted from markdown to MAML XML for code check-in and use at the command line.

I have a different message, let's talk about it in private.

PR #8540, opened 2021-11-02
Update Set-AutoSensitivityLabelPolicy.md

tiffanyyumsft:
Need the -startsimulation $true piece to start simulation

PR #8617, opened 2021-11-17
Update Set-AutoSensitivityLabelPolicy.md

chrisda:
Recreation of #8540


garycentric:
@chrisda
For angle brackets to be displayed in the published page, they need to be preceded with a backslash to tell the Markdown processor to not interpret the contents as HTML tags; for example, we need to use `\<placeholder\>` instead of `<placeholder>`.


 That's why this PR has instances of  **[Suggestion: disallowed-html-tag]**. Though these are reported as suggestions, they can have a significant effect on content. For example, this code:

```
### -AddOneDriveLocation
The AddOneDriveLocation parameter specifies the OneDrive for Business sites to add to the list of included sites when you aren't using the value All for the OneDriveLocation parameter. You identify the site by its URL value.

To enter multiple values, use the following syntax: <value1>,<value2>,...<valueX>. If the values contain spaces or otherwise require quotation marks, use the following syntax: "<value1>","<value2>",..."<valueX>".
```

Is rendered like this:

image

PR #8624, opened 2021-11-19
Update Set-Place.md

Luspin:
After some testing, we found that the entries for the GeoCoordinates parameter need to use period separators, instead of commas, e.g.:
Set-Place: Cannot process argument transformation on parameter 'GeoCoordinates'. Cannot convert value "48,643546;-122,12316" to type "Microsoft.Exchange.Data.GeoCoordinates". Error: "The latitude specified in '48,643546;-122,12316' is invalid. Latitude must be a number between -90.0 and 90.0."


chrisda:
@Luspin, see PR [6646](https://github.com/MicrosoftDocs/office-docs-powershell/pull/6646/files), where they changed periods to commas based on testing, and you're asking to change from commas to periods based on testing.

Could this be a localization/regional settings thing?

PR #8708, opened 2021-12-12
Update Get-Recipient.md

kellybos:
There are some discrepancies in this article that are leading to customer escalations -

1. Filter parameter uses OPATH filters instead of LDAP. There is a reference to filterable properties in Exchange server content which is not accurate for EXO. There are some properties like ExternalDirectoryObjectId which cannot be used and others like ExchangeGUID that don't accept wildcards. There should be reference to this article for most up-to-date supported attributes for EXO - https://docs.microsoft.com/en-us/powershell/exchange/filters-v2?view=exchange-ps.

2. RecipientPreviewFilter parameter doesn't use OPATH, it uses LDAP. The information provided in this parameter section is incorrect. Query must be in different format (unable to find official docs content on LDAP filtering), but an example would be - 

```
(&(mail=*)(msExchArchiveGUID=*))
```

LDAP uses the ActiveDirectory name of the property and different operators. Also allows more wildcard options than OPATH.


chrisda:
@kellybos, be advised that the Filter parameter with this exact same description exists in 29 other Exchange cmdlet reference topics, so I used my edited version of your update in those other descriptions.

PR #8713, opened 2021-12-14
Update Set-Mailbox.md

mprentice22153:
Typo in the first line.  Removed Exchange 2016 and Exchange 2019 since it is only available for Exchange Online.  Confirmed it does not work in Exch 2016 with a recent update applied.

### DIFF
--- a/exchange/exchange-ps/exchange/Set-Mailbox.md
+++ b/exchange/exchange-ps/exchange/Set-Mailbox.md
@@ -1308,7 +1308,7 @@ Accept wildcard characters: False
 ```
 
 ### -AccountDisabled
-This parameter is functional only in the cloud-based service.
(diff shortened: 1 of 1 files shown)


chrisda:
@mprentice22153, I fixed the typo ("o" to "or"). Availability and functionality are two different things. I confirmed that the parameter is available in Exchange 2016 and 2019 (the Applicable: metadata tag) so I'm adding that back. Whether or not the parameter actually works in Exchange 2016 or 2019 is a completely different story, hence the "This parameter is **available or functional** only in the cloud-based service** text, which I added back, since it's used everywhere for situations like this. If it was truly a cloud only parameter in a cloud/on-prem cmdlet, Applicable: would be Exchange Online, and the description would contain "This parameter is **available** only in the cloud-based service."

PR #8819, opened 2022-01-17
Update Get-Recipient.md

abhisk11:


get-itips (commented):
Hello @abhisk11 
The parameter sets already show a relation between `-IncludeSoftDeletedRecipients` and `-Identity` so I am not sure...

Also, the build is throwing errors because the note is not expected to be there.

dariomws:
Agree with @get-itips and...

If you have a Microsoft alias, you can link your accounts in GitHub so that a blue "Microsoft submitter" label appears to any PR's or issue you submit. Guidance on it here: https://docs.opensource.microsoft.com/tools/github/index.html

Here is the site you can use to see the mappings between GitHub and Microsoft aliases: https://repos.opensource.microsoft.com/people?q=abhisk11

PR #8839, opened 2022-01-21
PowerShell 7 modifications & removal of Ex2013

trudolf-msft:
PowerShell community SME answered this:

Yeah, that "encoding" was removed in PS7.
You can get the bytes content using the "-AsByteStream" parameter.
If you want something that works reliably over all PS versions, you can use the System.IO.File static class:
$r = [System.IO.File]::ReadAllBytes('C:\data\test.eml)
(Has better performance anyway)


JohanFreelancer9 (approved):

chrisda:
@trudolf-msft, like it or not, Test-Message is available in Exchange 2013, so we can't just remove the cmdlet/parameter applicable/Applicable values simply because we don't like them; and we don't like them, but I doubt there will ever be code work to hide/remove the cmdlet.

Appreciate the System.IO.File update. There are a TON of parameters in other cmdlets that are of parameter type = Byte[] and use the 'bad' `([Byte[]](Get-Content -Encoding Byte -Path...` syntax, so I'll find and replace those, too.

PR #9042, opened 2022-03-03
Update Get-ComplianceSearchAction.md

golinku:
\ is not part of valid syntax for passing ComplianceJobName.


chrisda:
@golinku, good catch. Underscore is a markdown character that's uses for italic formatting, so it's always best to escape it with a backslash to avoid confusion. But, if you surround the text in code, you need to remember to remove the backslash. I found a couple of other instances in the topic that I also fixed in this pull request.

PR #9095, opened 2022-03-13
Fixed bullets in Power BI entries

paulinbar:


paulinbar:
#sign-off
Hi Chris, I fixed a problem with the bullets in the Power BI entries and also clarified the text a little. The two files are ready to go. Thanks!
@chrisda
@adbenita

PR #9110, opened 2022-03-16
Update Get-MessageTrackingReport.md

KevinShaugh:
Removed "Exchange Online" from "applicable" - the synopsis clearly states it's only available in on-premises Exchange


chrisda:
@KevinShaugh, sorry; Get-MessageTrackingReport is 100% available in Exchange Online PowerShell as described. The applicable: (cmdlet) and Applicable: (parameter) values in the cmdlet reference topics are calculated and need to match availability reality, regardless of intent.

If/when (emphasis on if) the cmdlet is truly removed/hidden from customers (RBAC, flighting, etc.), my next sweep will catch it, and I'll remove the Exchange Online values for applicable:/Applicable:.

So, I'm closing this PR.

KevinShaugh:
@chrisda if so then the synopsis shouldn't say it's only available in exchange on-prem.

chrisda:
Changed handled globally everywhere per PR #9126, so closing

PR #9292, opened 2022-04-18
Kccross purview rebrand

KCCross:


KCCross:
Thomas requested permission to play with the contents of this branch for troubleshooting purposes. I moved the real content elsewhere. The fix to the 2818 files was to turn off docs build real-time validation in the docs validation extension in vs code.

PR #9333, opened 2022-04-22
Update New-UnifiedAuditLogRetentionPolicy.md

brucehuy:
@lindalu-MSFT tagging per Mark

PR #9336, opened 2022-04-25
Update New-UnifiedAuditLogRetentionPolicy.md

brucehuy:
@LauraWi @lindalu-MSFT we need to urgently revert this back as the documentation as originally intended should be the correct behavior and we should treat this as a code bug.

### DIFF
--- a/exchange/exchange-ps/exchange/New-UnifiedAuditLogRetentionPolicy.md
+++ b/exchange/exchange-ps/exchange/New-UnifiedAuditLogRetentionPolicy.md
@@ -71,7 +71,7 @@ Accept wildcard characters: False
 ```
 
 ### -Priority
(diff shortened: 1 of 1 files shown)


brucehuy:
@Chris Davis ***@***.***> – can we approve this ASAP?  We need to urgently revert this back as the documentation as originally intended should be the correct behavior and we should treat this as a code bug instead of a doc bug.

JohanFreelancer9 (commented):
@brucehuy, this appears to be a duplicate of PR #9333.

lindalu-MSFT:
I don’t have permissions in this repo, sorry I can’t help.

Get Outlook for iOS<https://aka.ms/o0ukef>

chrisda:
@brucehuy, always be mindful that 99% of the time, a change to a parameter in a New- cmdlet is also required in the corresponding Set- cmdlet.

I took care of this in PR #9340

PR #9436, opened 2022-05-17
Update Set-CsAutoAttendant.md

RuiPereiraTabares:
Looks like some versions of PowerShell (e.g. 5.1) will convert CallHandlingAssociations to array, while others (e.g. 7.0) will convert them to list. Runing:
$autoAttendant.CallHandlingAssociations.GetType().Name
5.1 => CallHandlingAssociation[] 
7.0 => ArrayList

the examples used will work only in 7.0 and not in 5.1


yogkumgit:
@JohanFreelancer9

dariomws (commented):
Hi @RuiPereiraTabares,

I cannot see the difference:
image

image

dariomws:
New-CsAutoAttendantCallHandlingAssociation also creates a CallHandlingAssociation object:

image

image

image

get-itips:
@RuiPereiraTabares 
Last time I asked, Teams PowerShell **is not** officially supported in Core/7, so I think this should not be added at the moment.

@pbafna03 for confirmation

get-itips (commented):
@RuiPereiraTabares
Last time I asked, Teams PowerShell is not officially supported in Core/7, so I think this should not be added at the moment.

@pbafna03 for confirmation

RuiPereiraTabares:
@get-itips  the thing is that our example 3 will not work on powershell 5, so we need to either edit the example or mention that example 3 will only work with powershell 7

pbafna03:
Yeah instead of this, I think better would be to tweak the example which doesn't work for PS 5. Also looping in @CLYVR .

CLYVR:
> Yeah instead of this, I think better would be to tweak the example which doesn't work for PS 5. Also looping in @CLYVR .

Agreed - if Powershell 7 is not officially supported yet then we shouldn't be showing a powershell 7 example and should change it so it works with powershell 5

RuiPereiraTabares:
well them this is the example 3 edititon for removing an callhandelingassociation

$autoAttendant.CallHandlingAssociations = $autoAttendant.CallHandlingAssociations | Where-Object{$_.ScheduleId -ne "0117420f-e04b-4b31-b370-9080ed6c4692"}

get-itips:
@CLYVR @pbafna03 Thanks, I think Example 3 was working until a recent version of Teams PS that made it stop working, 
@RuiPereiraTabares can you edit your PR to use PS5 version of the example?
Thanks

RuiPereiraTabares:
@get-itips done

JohanFreelancer9 (approved):

dariomws (approved):

get-itips (approved):

yogkumgit:
@tiburd Please merge this PR. Thanks

PR #9438, opened 2022-05-17
Updates to New/Set-CAR due to customer escalations.

briandaymsft:
Recent customer escalations have identified a need for clarification in our Client Access Rule cmdlet documentation to ensure rules are configured to get the desired results. We are clarifying how to configure country or region, user recipient filters, and providing additional examples of each.

Some minor markdown formatting suggestions from VS Code are also included.


briandaymsft:
(re-opening PR per recommendation in build error during first attempt) 

Recent customer escalations have identified a need for clarification in our Client Access Rule cmdlet documentation to ensure rules are configured to get the desired results. We are clarifying how to configure country or region, user recipient filters, and providing additional examples of each.

Some minor markdown formatting suggestions from VS Code are also included.

chrisda:
@briandaymsft, I appreciate the effort, but a couple of issues:

- Markdown lint 'issues' (i.e., missing carriage returns, some unfenced PowerShell code blocks, etc.) in cmdlet reference topics are a known issue and not something to be 'fixed'. The platyPS utility creates the topics with those 'errors' which don't affect publishing. Until platyPS itself changes, I'd rather have them all consistently 'wrong' rather than a few manually made right. A big global automated way to fix them all might happen in the future (which onesy-twosey fixes would interfere with).
- Tables and other advanced formatting (even nested lists!) are not allowed in cmdlet reference topics. Remember, these same topic files are converted to XML for use at the command line (Get-Help <Cmdlet>), where a table wouldn't work at all. I converted the information in your table back into a couple of lists. Same information, just not as pretty.

PR #9495, opened 2022-06-01
Delete Test-CsOnlineLisCivicAddress.md

JensTrier:
This cmdlet has been removed from the Teams PS Module.

I'm not clear if we have a process for leaving the PS cmdlet page, but creating this PR to remove the page.

Please let me know if this is against the process.


tseward:
@JensTrier typically we have a note that this cmdlet has been removed as of N version of the module. This way people using downlevel modules who still have the cmdlet have a reference available to them.

JensTrier:
Ok, makes sense. I'll cancel this PR and create a new one where I change the text.

PR #9501, opened 2022-06-02
Update Get-SafeLinksPolicy.md

msbemba:
Updated the document by removing isenabled and isdefault as they are no longer available in the latest module


msbemba:
Issue#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/8636

PR #9596, opened 2022-06-28
Update Set-SafeLinksRule.md

joscurtis:
Remove mention of dynamic distribution groups as these are not supported at this time.


chrisda:
@joscurtis, I happened to make these same changes yesterday, so your updates are conflicting with mine that were already merged. So, I'm closing your PR.

PR #9629, opened 2022-07-06
Update default value for AllowEngagementReport

robertpp:
The default value for AllowEngagementReport is Enabled if no value is specified when the Teams meeting policy is created from PowerShell. This is the opposite of what happens when a new meeting policy is created in Teams admin center.

image

### DIFF
--- a/skype/skype-ps/skype/New-CsTeamsMeetingPolicy.md
+++ b/skype/skype-ps/skype/New-CsTeamsMeetingPolicy.md
@@ -592,7 +592,7 @@ Aliases:
 
 Required: False
 Position: Named
-Default value: None
+Default value: Enabled
 Accept pipeline input: False
 Accept wildcard characters: False
 ```


yogkumgit:
@JohanFreelancer9

yogkumgit:
@dariomws @get-itips @tseward

get-itips:
Hello @robertpp Robert
Actually, the new policy values will depend of the current values in Global policy at the time of the cmdlet run:

image

JohanFreelancer9 (approved):
Copy review.

yogkumgit:
@robertpp

robertpp:
Hi @get-itips , thanks for the information, I wasn't aware of it. I confirm seeing the same behavior in our tenant, but totally missed the fact that the Global was no longer set to the default. This should be reflected in the documentation, as it is confusing, especially since the TAC behavior is different.

I would suggest the following changes:
a. Include a mention in the description: The default value of AllowEngagementReport in the Global policy is Disabled.
b. Set 'Default value:' to 'Matches (value in) Global policy'.

yogkumgit:
@robertpp Please comment.

robertpp:
Thank you for the comment @get-itips . I understand now the reason for _None_. The [documentation](https://docs.microsoft.com/en-us/microsoftteams/teams-analytics-and-reports/meeting-attendance-report) says the meeting attendance reports download is off by default, which to me it means the default value in the Global policy is turned off. 

image

The confusion starts when AllowEngagementReport in the Global policy is turned on. To prevent false expectations, it might be worth indicating the dependency on the value of AllowEngagementReport in the Global policy for subsequent policies.

get-itips:
It seems that sentence is outdated as the [linked article](https://docs.microsoft.com/en-us/microsoftteams/meeting-policies-in-teams-general#engagement-report) says

> This policy is **on by default** and allows your organizers to see who registered and attended the meetings and webinars they set up.

On new tenants, in the global policy, it appears to be on by default.

So I think the article requiring an edit is the one you linked that says it is off by default.

About this one, a note can be added that the initial values of the new policy are copied from the Global one.

robertpp:
Hi @get-itips . Thank you for the feedback. I raised another item for the other article. It's surprising to learn that it is enabled by default in Global policy. The entire documentation, including internal, except the article you mentioned, indicates it is off by default for privacy reasons. I even have a case where the customer is complaining about the feature not being compliant with GDPR. That's how I got to raise these items. Your efforts and clarifications are much appreciated.

dariomws:
@get-itips @robertpp, it's enabled in my demo tenant, it should be the default value for the Global policy.

yogkumgit:
@get-itips @dariomws

get-itips:
@robertpp Robert, sorry, on my opinion, I was more inclined to leave "None" as value.

>  the general convention is using None when the value can not be exactly determined as it depends on Global policy/current settings in customer tenant.

> About this one, a note can be added that the initial values of the new policy are copied from the Global one.

@dariomws thoughts?

dariomws:
Agree with @get-itips.

yogkumgit:
@robertpp Please comment

yogkumgit:
@robertpp  How you wish to proceed with this PR. Please suggest.
CC: @tiburd

robertpp:
Hello @yogkumgit and @tiburd . Apologies for the slow follow up. I was on vacation the past 2 weeks. I agree with @get-itips' suggestion.

yogkumgit:
@robertpp Thank you and no worries. I am closing it.

PR #9678, opened 2022-07-14
Update Set-TransportConfig.md

MoibraM:
The (-DSNConversionMode) explanation 
Original:  PreserveDSNBody (This is the default value)
Updated: UseExchangeDSNs  (This is the default value)
@chrisda for your review


chrisda:
I did some testing on this. Here's what I found:

> The default value in Exchange 2010 and Exchange 2013 is UseExchangeDSNs. Otherwise, the default value is PreserveDSNBody.

PR #9746, opened 2022-08-02
Added information about UTC time format

Petri-X:
It could be confusing for user when s/he is filtering emails on local time format, but output is showing actually UTC time format.

There is also a tiny challenge, as if you use "get-date -adddays ()" for defining the start-/enddates you are looking for, then you are getting a bit different results.

But purpose of this was just clarify the UTC time format on the output.

### DIFF
--- a/exchange/exchange-ps/exchange/Get-MessageTrace.md
+++ b/exchange/exchange-ps/exchange/Get-MessageTrace.md
@@ -46,7 +46,7 @@ To search for message data that is greater than 10 days old, use the Start-Histo
 
 By default, this cmdlet returns a maximum of 1000 results, and will timeout on very large queries. If your query returns too many results, consider splitting it up using smaller StartDate and EndDate intervals.
 
-The time stamps on the output are in UTC time format. That might be different time format than you were using for the -StartDate and the -EndDate parameters.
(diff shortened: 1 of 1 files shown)


exchange/exchange-ps/exchange/Get-MessageTrace.md
```diff
@@ -46,6 +46,8 @@ To search for message data that is greater than 10 days old, use the Start-Histo
 
 By default, this cmdlet returns a maximum of 1000 results, and will timeout on very large queries. If your query returns too many results, consider splitting it up using smaller StartDate and EndDate intervals.
 
+The time stamps on the output are in UTC time format. That might be different time format than you were using for the -StartDate and the -EndDate parameters.
```
JohanFreelancer9, line 49:
```suggestion
The time stamps on the output are in UTC time format. That might be different from the time format that you used for the -StartDate and the -EndDate parameters.
```

JohanFreelancer9 (changes requested):
Copy review - @Petri-X, kindly commit the minor suggested changes. Many thanks.

Petri-X:
Big thanks @JohanFreelancer9 for your review.

JohanFreelancer9 (approved):

PR #10026, opened 2022-10-10
RemotePowerShellEnabled disabled for every user

Sartharon:
**Description**
Using `-or` comparison inside the Where-clause to exclude specified admins leads to having RemotePowerShell disabled for those users as well.
Example:
`admin1 -ne admin1 -or admin1 -ne admin2`
-> `$false -or $true`
-> `$true`

Using the `-and` comparison fixes this behavior, since neither admin1 nor (`-ne` + `-and`) admin2 should have their RemotePowerShell disabled.

**Real world tests**
- Default State:
image
- After running the suggested cmdlet with **administrator@lab.local** __and__ **admin2@lab.local** exempt
`Get-User | Where-Object {$_.UserPrincipalName -ne 'administrator@lab.local' -or $_.UserPrincipalName -ne 'admin2@lab.local'} | Set-USer -RemotePowerShellEnabled $false`
image
- Using the corrected cmdlet with an `-and` comparison:
image


JohanFreelancer9 (approved):

chrisda:
Already fixed yesterday #10027, thus now this PR results in merge conflicts.

Closing.

PR #10056, opened 2022-10-18
Update New-DlpCompliancePolicy.md

ybhargav1995:


JohanFreelancer9 (approved):

chrisda:
@ybhargav1995, you're saying that non-mail enabled groups (Azure?) are OK to use with ExchangeSenderMemberOf[Exception] in New- and also presumably in Set-? That seems a bit odd; how can those groups receive mail?

ybhargav1995:
Hi Chris,

They need not send the mail from the that group account. They could be sending it from their personal accounts, since the predicate is Exchange sender member of which means if they are part of this group the policy must apply.

Thanks,
Bhargav

PR #10096, opened 2022-10-28
Remove -Confirm

m2rinch:
We do not support this command anymore documentation needs to be updated to match


JohanFreelancer9 (approved):

m2rinch:
Can we please get this merged asap, it is customer impacting

chrisda:
I confirmed that the Confirm parameter is still available in my plain (no flighting, no SDF) M365 E5 tenant.

Because the parameter is still available, I'm adding it back, but with the description "This parameter is reserved for internal Microsoft use."

Over my years of doing this, I've found it's better to leave available parameters in the cmdlet reference topics, but describe them as nice versions of "this doesn't work; don't use it", which "reserved for internal Microsoft use" does very well.

If/when the parameter is actually removed from visibility for customers, we'll remove it from content.

PR #10156, opened 2022-11-17
Changes for SMimeSuppressNameChecksEnabled switch

anepu:
For the SMimeSuppressNameChecksEnabled switch can we modify the documentation as currently it says This parameter is reserved for internal Microsoft use. However, this is usable in PowerShell for tenants and even more this was approved/implemented as DCR somewhere in April 2022.

Tested this in test tenant and the switch works as intended (per info in the DCR)


chrisda:
@anepu, removing the 'reserved for internal Microsoft use' is great, but then an actual description for the parameter is required. I bluffed my way through one. Feel free to modify it.

PR #10242, opened 2022-12-13
Update Add-RoleGroupMember.md

MSFTAnthonyLiu:
Confirmed with the engineering team who manages the UI in the Microsoft Purview Portal, the current legacy picker to select the members has no ability to select Security Groups and will be retired in the following months. The new picker is able to select Security Group and will replace the old one. 
Hence, we need to update this article for now to inform customer and guide them to use the cmdlet.

### DIFF
--- a/exchange/exchange-ps/exchange/Add-RoleGroupMember.md
+++ b/exchange/exchange-ps/exchange/Add-RoleGroupMember.md
@@ -96,6 +96,9 @@ The Member parameter specifies who you want to add to the role group. You can sp
 - Mail-enabled security groups (don't use in Security & Compliance PowerShell)
 - Security groups (on-premises Exchange only)
 
+Kindly note that the Security Groups/Mail-enabled security groups cannot be added as a member of the role group in the Microsoft Purview Portal, please add them via PowerShell with this parameter.
+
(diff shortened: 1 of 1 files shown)


exchange/exchange-ps/exchange/Add-RoleGroupMember.md
```diff
@@ -96,6 +96,9 @@ The Member parameter specifies who you want to add to the role group. You can sp
 - Mail-enabled security groups (don't use in Security & Compliance PowerShell)
 - Security groups (on-premises Exchange only)
 
+Kindly note that the Security Groups/Mail-enabled security groups cannot be added as a member of the role group in the Microsoft Purview Portal, please add them via PowerShell with this parameter.
+
```
JohanFreelancer9, line 100:
```suggestion
Note that you cannot add security groups and mail-enabled security groups as members of the role group in the Microsoft Purview portal. You should add them via PowerShell with this parameter.
```

JohanFreelancer9 (changes requested):
Copy review - @MSFTAnthonyLiu, kindly commit the suggested changes. Many thanks.

PR #10288, opened 2022-12-31
Update app-only-auth-powershell-v2.md

Pallavi-mp:
Modified the admin consent part for multi-tenant applications, since the interactive consent does not work with our module. Admin consent URL should be used to grant consent.


Pallavi-mp:
@chrisda please take a look at this change

PR #10318, opened 2023-01-10
Update Set-CalendarProcessing.md

robchantler:


robchantler (commented):
Change "the" entire series to "an" entire series as the setting affects Individual meetings and Series.

PR #10509, opened 2023-02-24
Update Remove-VivaModuleFeaturePolicy.md

edwin-king:
@chrisda, making a change to this file.


chrisda:
@edwin-king, let's keep the parameters in alpha order in the SYNTAX and PARAMETERS section (Required: True first, then Required: Optional). Any order is fine in the examples.

If the Position values for the parameters were numbered, we could sort them on that.

PR #10679, opened 2023-03-30
Update New-TransportRule -SenderAddressLocation section

m-nestorovic:
The documentation is incorrect.
When Header is used during the evaluation of the rule, only property FROM from message header is used. Sender and Reply-To properties are not used.
Confirmation of this can be found in the ADO WI.

It should state:
**Header**: Only examine senders in the message headers (**From** field). This is the default value.

The ADO Work Item to correct this mistake can be found here https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/3510129

### DIFF
--- a/exchange/exchange-ps/exchange/New-TransportRule.md
+++ b/exchange/exchange-ps/exchange/New-TransportRule.md
@@ -4415,11 +4415,11 @@ Accept wildcard characters: False
 ### -SenderAddressLocation
 The SenderAddressLocation parameter specifies where to look for sender addresses in conditions and exceptions that examine sender email addresses. Valid values are:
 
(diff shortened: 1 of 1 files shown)


chrisda:
@m-nestorovic, a couple of suggestions:

- These pull requests or public, so don't include links to support cases (the support case number alone is fine)
- Be aware that cmdlet reference topics are single-sources for on-prem Exchange AND Exchange Online. I presume this issue affects both on-prem and online.

m-nestorovic:
@chrisda Noted.
Will not post ADO cases.
And will check for the On-Prem, and propose change if needed.

Thanks for the suggestions. Much Appreciated.

PR #10719, opened 2023-04-11

ybhargav1995:


chrisda:
@ybhargav1995, I'm OK with a lot of your changes, but some of your changes were to boilerplate text that's used in other cmdlet descriptions (hundreds in the case of Confirm), so I changed some back.

PR #10803, opened 2023-05-04
Update New-TransportRule

m-nestorovic:
Added a fact that "If the rule is created without any condition and exception, that rule will match for all messages."

Such rule has huge impact and often lead to the catastrophic scenario for the customer if not set properly.

Work Item ID: 3103951

### DIFF
--- a/exchange/exchange-ps/exchange/New-TransportRule.md
+++ b/exchange/exchange-ps/exchange/New-TransportRule.md
@@ -16,7 +16,7 @@ This cmdlet is available in on-premises Exchange and in the cloud-based service.
 
 Use the New-TransportRule cmdlet to create transport rules (mail flow rules) in your organization.
 
-**Note**: The action of a rule without conditions or exception is applied to all messages, which could have unintended consequences. For example, if the rule action deletes messages, the might delete all inbound and outbound messages for the entire organization.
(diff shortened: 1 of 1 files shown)


exchange/exchange-ps/exchange/New-TransportRule.md
```diff
@@ -16,6 +16,8 @@ This cmdlet is available in on-premises Exchange and in the cloud-based service.
 
 Use the New-TransportRule cmdlet to create transport rules (mail flow rules) in your organization.
 
+**Note**: The action of a rule without conditions or exception is applied to all messages, which could have unintended consequences. For example, if the rule action deletes messages, the rule without conditions or exceptions might delete all inbound and outbound messages for the entire organization.
```
JohanFreelancer9, line 19:
```suggestion
**Note**: The action of a rule without conditions or exceptions is applied to all messages, which could have unintended consequences. For example, if the rule action deletes messages, the rule without conditions or exceptions might delete all inbound and outbound messages for the entire organization.
```

JohanFreelancer9 (changes requested):
Copy review - @m-nestorovic, Kindly commit the minor suggested change. Many thanks.

JohanFreelancer9 (approved):

PR #10804, opened 2023-05-04
Update Set-TransportRule

m-nestorovic:
Highlighted the fact that "If the rule is created without any condition and exception, that rule will match for all messages."

Such rule has huge impact and often lead to the catastrophic scenario for the customer if not set properly.

Work Item ID: 3103951

### DIFF
--- a/exchange/exchange-ps/exchange/Set-TransportRule.md
+++ b/exchange/exchange-ps/exchange/Set-TransportRule.md
@@ -16,8 +16,7 @@ This cmdlet is available in on-premises Exchange and in the cloud-based service.
 
 Use the Set-TransportRule cmdlet to modify existing transport rules (mail flow rules) in your organization.
 
-> [!NOTE]
-> If you delete all conditions and exceptions from a rule, the rule action is applied to all messages. This can have unintended consequences. For example, if the rule action is to delete the message, removing the conditions and exceptions could cause the rule to delete all inbound and outbound messages for the entire organization.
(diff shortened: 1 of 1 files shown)


exchange/exchange-ps/exchange/Set-TransportRule.md
```diff
@@ -16,7 +16,7 @@ This cmdlet is available in on-premises Exchange and in the cloud-based service.
 
 Use the Set-TransportRule cmdlet to modify existing transport rules (mail flow rules) in your organization.
 
-If you delete all conditions and exceptions from a rule, the rule action is applied to all messages. This can have unintended consequences. For example, if the rule action is to delete the message, removing the conditions and exceptions could cause the rule to delete all inbound and outbound messages for the entire organization.
+**Note**: The action of a rule without conditions or exception is applied to all messages, which could have unintended consequences. For example, if the rule action deletes messages, the rule without conditions or exceptions might delete all inbound and outbound messages for the entire organization.
```
JohanFreelancer9, line 19:
```suggestion
**Note**: The action of a rule without conditions or exceptions is applied to all messages, which could have unintended consequences. For example, if the rule action deletes messages, the rule without conditions or exceptions might delete all inbound and outbound messages for the entire organization.
```

JohanFreelancer9 (changes requested):
Copy review - @m-nestorovic, Kindly commit the minor suggested change. Many thanks.

JohanFreelancer9 (approved):

PR #10836, opened 2023-05-11
Update New-TransportRule.md

MithunRathinam:
“Applies to:” with ExO and ExO Online Protection
is misleading/confusing because it’s the opposite message than the statement at the beginning that SCL-Exceptions DO NOT apply in ExO.


chrisda:
@MithunRathinam, same comment and actions as #10837.

PR #10837, opened 2023-05-11
Update Set-TransportRule.md

MithunRathinam:
“Applies to:” with ExO and ExO Online Protection
is misleading/confusing because it’s the opposite message than the statement at the beginning that SCL-Exceptions DO NOT apply in ExO.

### DIFF
--- a/exchange/exchange-ps/exchange/Set-TransportRule.md
+++ b/exchange/exchange-ps/exchange/Set-TransportRule.md
@@ -4304,7 +4304,7 @@ Accept wildcard characters: False
 ```
 
 ### -SCLOver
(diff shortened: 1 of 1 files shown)


MithunRathinam:
@chrisda , to merge PR

chrisda:
@MithunRathinam, that's all good, but that parameter is available In ExO/EOP PowerShell. Parameter 'applicable' values must stay technically accurate. If/when SCLOver is made not available in ExO/EOP PowerShell, we can remove those 'applicable' values.

For parameters in this state, we use the text description to explain (and this kind of thing happens a lot):

> **Note**: This parameter is functional only in on-premises Exchange.

I modified your PR accordingly.

PR #10913, opened 2023-06-06

ybhargav1995:


chrisda:
@ybhargav1995, do this same changes also need to be made to the [ExceptIf]DocumentIsPasswordProtected parameter descriptions in New/Set-AutoSensitivityLabelRule?

ybhargav1995:
> 

@chrisda - yes it is required

PR #10965, opened 2023-06-20
Update Set-MailboxJunkEmailConfiguration.md

MithunRathinam:
Removed below old lines which is no more the case.

You can only use this cmdlet on a mailbox that's been opened in Outlook (in Cached Exchange mode) or Outlook on the web. If the mailbox hasn't been opened, you'll receive the error: "The Junk Email configuration couldn't be set. The user needs to sign in to Outlook Web App before they can modify their Safe Senders and Recipients or Blocked Senders lists." If you want to suppress this error for bulk operations, you can add -ErrorAction SilentlyContinue to the end of the command.

### DIFF
--- a/exchange/exchange-ps/exchange/Set-MailboxJunkEmailConfiguration.md
+++ b/exchange/exchange-ps/exchange/Set-MailboxJunkEmailConfiguration.md
@@ -16,8 +16,6 @@ This cmdlet is available in on-premises Exchange and in the cloud-based service.
 
 Use the Set-MailboxJunkEmailConfiguration cmdlet to configure the junk email settings on mailboxes.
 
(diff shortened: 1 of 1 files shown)


MithunRathinam:
@chrisda To merge PR

chrisda:
@MithunRathinam, is this true for both on-prem Exchange and Exchange Online? I'll bet your change is accurate for ExO, but not for on-prem Exchange Server.

MithunRathinam:
> @MithunRathinam, is this true for both on-prem Exchange and Exchange Online? I'll bet your change is accurate for ExO, but not for on-prem Exchange Server.

@chrisda , Tested for both onpremise and EXO Mailboxes . 
1) In EXO, default was isenabled : false , we need to enable- It works without the mailbox being opened in OWA/Outlook
2) In Onpremise Exchange , default was isenabled: True. 

In both the Onpremise/EXO, I am able to set any parameters such as "Trustedsenderanddomains" etch without mailbox being opened in OWA/Outlook for the first time. so This applied to both Exchange onpremise and EXO

PR #11060, opened 2023-07-12
DLP updates to Transport Rule cmdlets

m-nestorovic:
Removed -DlpPolicy parameter since it is depricated.

Work Item ID in the ADO: 3768407


JohanFreelancer9 (approved):

chrisda:
@m-nestorovic, I'm sure that's true for Exchange Online, but I doubt it's true for Exchange Server. We need to be careful as Exchange cmdlet reference topics are single sourced for the cloud and multiple versions of Exchange Server.

m-nestorovic:
@chrisda you are absolutely right. This is true only for the Exchange Online, but not for the Exchange Server.
Can you help me align this PR to present that information correctly?

m-nestorovic:
@chrisda I have reviewed PR it is good to be merged.
We can merge it as soon as possible, since the rollout with this change is in progress already.

PR #11061, opened 2023-07-12
Update Set-TransportRule.md

m-nestorovic:
-DlpPolicy parameter has been deprecated, and we want to make clear statement about that.

Work Item ID in the ADO: 3768407

### DIFF
--- a/exchange/exchange-ps/exchange/Set-TransportRule.md
+++ b/exchange/exchange-ps/exchange/Set-TransportRule.md
@@ -1316,8 +1316,8 @@ Accept wildcard characters: False
 
 ### -DlpPolicy
 
-Existing Transport rules with this property are no longer being evaluated or executed because it is linked to an Exchange DLP policy. 
-Please migrate the policies to the M365 Compliance Center by using the migration wizard.
+Existing transport rules with this property are no longer evaluated or executed because it is linked to an Exchange DLP policy. 
+You should migrate the policies to the Microsoft 365 compliance center by using the migration step-by-step guide.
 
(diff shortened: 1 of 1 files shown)


exchange/exchange-ps/exchange/Set-TransportRule.md
```diff
@@ -1315,6 +1315,10 @@ Accept wildcard characters: False
 ```
 
 ### -DlpPolicy
+
+Existing Transport rules with this property are no longer being evaluated or executed because it is linked to an Exchange DLP policy. 
+Please migrate the policies to the M365 Compliance Center by using the migration wizard.
```
JohanFreelancer9, line 1320:
```suggestion
Existing transport rules with this property are no longer evaluated or executed because it is linked to an Exchange DLP policy. 
You should migrate the policies to the Microsoft 365 compliance center by using the migration step-by-step guide.
```

JohanFreelancer9 (changes requested):
Copy review - @m-nestorovic, kindly commit the suggested changes. For information, the latest guidance is not to use the term "wizard" in Microsoft docs but rather "step-by-step guide". Many thanks.

JohanFreelancer9 (approved):

chrisda:
@m-nestorovic, I'm sure that's true for Exchange Online, but I doubt it's true for Exchange Server. We need to be careful as Exchange cmdlet reference topics are single sourced for the cloud and multiple versions of Exchange Server.

If that's true, then we need to switch the old-style DLP parameters from availability in the cloud and on-prem to on-prem only.

m-nestorovic:
@chrisda Same as for another PR for New-TransportRule.
You are absolutely right. This is true only for the Exchange Online, but not for the Exchange Server.
Can you help me align this PR to present that information correctly?

chrisda:
Closing this as a duplicate of #1160. Will make all required changes there.

PR #11115, opened 2023-07-27
Update Get-InformationBarrierRecipientStatus.md

rasrivas-ms:
Add information about non-Legacy modes and alternatives to use if you are in a non-Legacy mode.


rasrivas-ms:
@chrisda Can you please review this change?

PR #11147, opened 2023-08-08
Update Set-PolicyConfig.md

rasrivas-ms:
- Added more information about IBMode in SetPolicyConfig
- Removed Legacy from set of valid values since you can't revert back to Legacy after switching to Single or Multi Segment modes.


rasrivas-ms:
@chrisda Please review and merge.

Issue #8537, opened 2021-10-31
Set-MailboxMessageConfiguration has no effect


leewisestamp:
We have been successfully updating OWA signatures using `Set-MailboxMessageConfiguration` for the past 6+ months. Suddenly, and with no change on our part, `Set-MailboxMessageConfiguration` has started to have no effect. Our business heavily depends on this cmdlet and we would really appreciate your feedback/guidance. Someone else posted the same issue on SO here: https://stackoverflow.com/q/69764573/152630.

Example (that previously worked):
```
Set-MailboxMessageConfiguration -Identity "joe@domain.com" -SignatureHTML "My new sig" -AutoAddSignature $True -AutoAddSignatureOnReply $True
```

Interestingly, after manually running `Set-MailboxMessageConfiguration` in a local PowerShell, we see the correct result in `Get-MailboxMessageConfiguration` (https://www.screencast.com/t/X9i4BY1jSb) but NOT in the OWA UI. Please advise.

leewisestamp:
Any guidance on the matter would be greatly appreciated

chrisda:
@leewisestamp, the best way for you to get an answer is to open a support ticket. The fact that this feature worked fine and then suddenly stopped working indicates some kind of problem that's beyond documentation.

Report back here with your findings.

leewisestamp:
@chrisda Obviously, we did that in parallel to opening this documentation ticket. So far Support has no answers other than a shaky workaround that will effectively ruin our product.

And I disagree that this is beyond documentation because if the cmdlet behavior has changed (e.g. to support multiple signatures in OWA) without that being documented (e.g. by adding a new parameter to the cmdlet to specify which signature should be updated), then that is a gap in documentation.

chrisda:
@leewisestamp, the issue is determining whether this is a deliberate feature change or some transitory problem that's going to get fixed. The answer to those questions determines whether we update documentation or not.

So is your ticket still open? Are they looking into it and gave you the lame workaround as a stop-gap until they can find out what's going on? Or did they say "that's the way it is" and close your ticket? And are you saying that there are new undocumented parameters/behavior that need to be documented?

leewisestamp:
@chrisda I don't work for MSFT so your team is far better qualified to determine if this deliberate or a bug.

Our ticket is still open with effectively no progress for 2 days other than the lame workaround and "we're looking into it".

[...]

PR #11213, opened 2023-08-29
Update Set-MailboxMessageConfiguration.md

Inderjeetsinghjaggi:
As per a case with MS TrackingID#2308240030000032, below switch doesn't work now and they are deprecated for Office 365. If they are deprecated, they should be removed from the article or should mention it only works for users with no signature and will only work for On prem exchange servers
 [-SignatureHtml <String>]
   [-SignatureText <String>]
   [-SignatureTextOnMobile <String>]


Inderjeetsinghjaggi:
@chrisda do we have any update on when the article will be updated with correct information

chrisda:
@Inderjeetsinghjaggi, this whole thing is a huge can of worms. See this still open issue from 2021: [#8537](https://github.com/MicrosoftDocs/office-docs-powershell/issues/8537).

Those parameters are still available to me in my plain old M365 E5 tenant. Until they actually go away, I'm reluctant to make any changes.

chrisda:
Relates to Issue #11371 and Issue #8537

PR #11333, opened 2023-09-29
Update Set-CalendarProcessing.md

braalf:
Adding in information about Query-based groups not being supported


braalf:
@chrisda for approval

braalf (commented):
Change looks good

PR #11355, opened 2023-10-07
Update New-AntiPhishPolicy.md

codefusser:
including the append hashtable to update existing TargetedUsersToProtect list


chrisda:
@codefusser, even though I merged this, I realize now it doesn't make sense in the New- cmdlet. Preserving existing values makes sense only in the Set- context. So, I removed your new information from New- and put it in Set-.

PR #11371, opened 2023-10-13
Update Set-OrganizationConfig.md

RajRitwik16:
Added the information about the new cmdlet Set-OrganizationConfig -PostponeRoamingSignaturesUntilLater $true


chrisda:
@RajRitwik16, I did some edits. We should merge this PR once WW saturation reaches ~50% (too early is bad, too late is bad, but too early is worse). Also, no way should we give any sort of timeline for future updates, so I removed that language.

chrisda:
Helps with #8537

Issue #11323, opened 2023-09-28
Get-Team -DisplayName still returns partial matches, contrary to the documentation


PerePasqual:
Hi,

Yesterday I wrote a new comment both on 
https://github.com/MicrosoftDocs/office-docs-powershell/issues/7636 (closed)
and
https://github.com/MicrosoftDocs/office-docs-powershell/pull/7692 (closed),
but none of the team members seem to care about that.

Still neither the behaviour of Get-Team -DisplayName has been fixed or the documentation has been rewritten properly to describe what the parameter does.

The docs still reads:

"Filters to return teams **with a full match to the provided displayname**.". That's false.

And then it continues like:

"As displayname is not unique, this acts as a filter rather than an exact match. Note that this filter value is case-sensitive."

As displayname is not unique and we can have several teams with exactly the same name, that should be the only case when this command with this parameter could and should several results, but this is not an excuse for this acting "as a filter rather than an exact match". It could have been implemented not as a filter but (as wrongly stated in the docs) as an exact and in the case of not unique displayname the behaviour could and should be the same as it's now, returning several results but only for an exact match.

You want to return partial matches, then you should include the ability of using wildcards in the displayname, as "MyTeam%" or so; that's the standard behaviour elsewhere.

This buggy behaviour may potentially be affecting thousands of users that are not aware of it and are using the parameter in production environment and lead to al kind of unexpected behaviour, like happened in my case.

Not having the ability to get a team by its EXACT displayname renders the whole cmdlet unreliable / useless.

get-itips:
Thanks for your feedback, I am proposing a better description of the actual behavior.

PR #11407, opened 2023-10-24
Improve displayName description

get-itips:
closes https://github.com/MicrosoftDocs/office-docs-powershell/issues/11323

### DIFF
--- a/teams/teams-ps/teams/Get-Team.md
+++ b/teams/teams-ps/teams/Get-Team.md
@@ -126,7 +126,7 @@ Accept wildcard characters: False
 ```
 
 ### -DisplayName
-Filters to return teams with a full match to the provided displayname.  As displayname is not unique, this acts as a filter rather than an exact match. Note that this filter value is case-sensitive.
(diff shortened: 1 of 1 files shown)


teams/teams-ps/teams/Get-Team.md
```diff
@@ -126,7 +126,7 @@ Accept wildcard characters: False
 ```
 
 ### -DisplayName
-Filters to return teams with a full match to the provided displayname.  As displayname is not unique, this acts as a filter rather than an exact match. Note that this filter value is case-sensitive.
+Specify this parameter to return teams with the provided displayname as a filter.  As displayname is not unique, multiple values can be returned. Note that this filter value is case-sensitive.
```
JohanFreelancer9, line 129:
```suggestion
Specify this parameter to return teams with the provided display name as a filter.  As the display name is not unique, multiple values can be returned. Note that this filter value is case-sensitive.
```

JohanFreelancer9 (changes requested):
Copy review.

JohanFreelancer9 (approved):

msbemba (approved):

yogkumgit:
@tiburd Please merge this PR. Thanks

PR #11422, opened 2023-11-02
Update Get-SafeLinksDetailReport.md

joscurtis:
Adding value definitions for the Action values as this has led to customer confusion.


chrisda:
Reformatted from table to bulleted list (table formatting not allowed; these articles are also used for help at the command line).

PR #11474, opened 2023-11-16
Updated Get-DistributionGroupMember.md-ADO-8426457

Ashok-Lobo:


Ashok-Lobo:
@chrisda Could you please guide how to resolve this error.

ADO: https://office.visualstudio.com/MAX/_workitems/edit/8426457/

chrisda (approved):
Don't try to fix the lack of a carriage return between the H2s/H3s and text. All cmdlet reference topics suffer from this 'problem', because that's the way the platyPS utility currently creates them. I'd rather them all be the same for possible future global fix.

You can't just delete empty sections you don't like as that violates the schema. Nor can you simply put text anywhere you like (hence the massive cryptic validation errors). Typically, I would suggest adding that type of information in the DESCRIPTION section, except you can't use code blocks there (code inline is OK).Therefore, I just added it as a new example.

Ashok-Lobo:
@chrisda Thank you very much! Going forward, I will follow this guidelines for all cmdlet reference topics.

PR #11506, opened 2023-12-04
Update Get-CaseHoldRule.md

pjsfonseca69:
timeout issue when running only the command Get-CaseHoldRule without any parameter

### DIFF
--- a/exchange/exchange-ps/exchange/Get-CaseHoldRule.md
+++ b/exchange/exchange-ps/exchange/Get-CaseHoldRule.md
@@ -27,7 +27,7 @@ Get-CaseHoldRule [[-Identity] <ComplianceRuleIdParameter>]
 ```
 
 ## DESCRIPTION
-In large environments, running this cmdlet without any parametersresults in a timeout. As a workaround, you can run the following command:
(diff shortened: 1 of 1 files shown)


chrisda:
@pjsfonseca69, where does your command get the $org variable from? We should declare it or remove it.

exchange/exchange-ps/exchange/Get-CaseHoldRule.md
```diff
@@ -27,6 +27,10 @@ Get-CaseHoldRule [[-Identity] <ComplianceRuleIdParameter>]
 ```
 
 ## DESCRIPTION
+In large environments, running this cmdlet without any parametersresults in a timeout. As a workaround, you can run the following command:
```
JohanFreelancer9, line 30:
```suggestion
In large environments, running this cmdlet without any parameters results in a timeout. As a workaround, you can run the following command:
```

JohanFreelancer9 (changes requested):
Copy review.

pjsfonseca69:
> @pjsfonseca69, where does your command get the $org variable from? We should declare it or remove it.

What do you mean, we don't use any $org variable we just run Get-CaseHoldRule

PR #11511, opened 2023-12-05
Update New-TransportRule.md

MithunRathinam:
Added a note on AD attributes used in Transport rule


MithunRathinam:
@chrisda to merge PR

PR #11546, opened 2023-12-21
Update Start-ManagedFolderAssistant.md

Logonzal:
Add opening and closing square bracket in the syntax example.

### DIFF
--- a/exchange/exchange-ps/exchange/Start-ManagedFolderAssistant.md
+++ b/exchange/exchange-ps/exchange/Start-ManagedFolderAssistant.md
@@ -20,7 +20,7 @@ For information about the parameter sets in the Syntax section below, see [Excha
 
 ## SYNTAX
 
-### Default
+### Identity (Default)
 ```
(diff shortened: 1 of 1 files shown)


JohanFreelancer9 (approved):

chrisda:
@Logonzal, you're right, but strange things are going on in that cmdlet.

The [platyPS](https://github.com/PowerShell/platyPS) utility we use to dump cmdlets to markdown is detecting the HoldCleanup and StopHoldCleanUp switch parameters as `Required: True` in the HoldCleanupParameterSet and StopHoldCleanupParameterSet parameter sets.  When I checked Exchange on-prem and Online code, those parameters are definitely tagged as `Required: False`.

If a switch parameter is required, by definition it can't have brackets around it, because brackets = optional, which is likely why at some point I removed the brackets. Why I also didn't remove them from StopHoldCleanup is unclear.

Unclear what's going on (I've never known platyPS to make mistakes like that ?!?), but I used my audit from last month to modify your PR to make sure everything matches current reality, including adding square brackets to HoldCleanup and the other ramifications of `Required: False`.

PR #11548, opened 2023-12-29
Update Export-ActivityExplorerData.md

aamalhot:
Goal is to provide close to a full script instead of single step which seems to be creating confusion currently


aamalhot:
@chrisda  for your review

PR #11550, opened 2024-01-02
Fix a few typos

Blake-Madden:
Fix misspellings, a messed up sentence, and some article issues


chrisda:
@Blake-Madden, all good catches, but in the future, please don't mix Exchange article changes with Skype/Teams article changes. Different areas of responsibility and approval,

PR #11577, opened 2024-01-19
Updated-8670703

v-jmathew:
Max length value for 'Name' parameter updated.


chrisda:
@v-jmathew, I added this change to your original PR  #11751. So is this PR now duplicative?

v-jmathew:
@chrisda, I just checked PR #11751. It seems like I received a duplicate request for this change as it has already been covered in the previous PR. I'll close out this PR.

Thank you for the update!

PR #11602, opened 2024-02-02
Documentation approval for cmdlet 'Remove-CsUserLicenseGracePeriod'

dShyam01:
The cmdlet was implemented to aid the delay delicensing feature in Teams, which enables tenant admins to reverse an assigned plan deletion by introducing a configurable grace period, during which the licenses will still be active. In case the plan was deleted accidentally, the admin can reverse the same. If not, running this command removes the grace period of the assigned plan(s) against the specified user(s)/resource account(s), permanently deleting the plan(s).


dShyam01:
@microsoft-github-policy-service agree company="Microsoft"

PR #11615, opened 2024-02-07
Update Get-EXOMailboxFolderStatistics.md

paisvivekanand:
The 2 parameters IncludeAnalysis and DiagnosticInfo is not supported for external customers in Get-EXOMailboxFolderStatistics


chrisda:
Parameters are still available, even in v3.4.2-Preview1. IncludeAnalysis is already described as "reserved for internal Microsoft use".

Closing

PR #11617, opened 2024-02-08
Update Set-CsTeamsVoiceApplicationsPolicy.md

CLYVR:
Updating to remove note that CQ must allow opt-in/out in order for an auth user to perform this function.


CLYVR:
@get-itips @msbemba Please and thanks.

CLYVR:
@get-itips @msbemba @tiburd Please and thanks.

PR #11619, opened 2024-02-09
Update New-CsCallQueue.md

CLYVR:
Adding new *CallPriority parameters in order to publish new cmdlets.  Parameters are currently restricted to private preview customers.


CLYVR:
@get-itips @msbemba @tiburd Please and thanks!

PR #11709, opened 2024-04-07
Fix a few typos

Blake-Madden:


hirenshah1:
I am no longer owner for exchange.

PR #11710, opened 2024-04-08
Removing -whatif parameter which is not supported for theses cmdlets

williamlooney:
Removing -WhatIf parameter which is unsupported in this cmdlet


williamlooney:
Trying again

PR #11714, opened 2024-04-08
Removing -WhatIf parameter

williamlooney:
These cmdlets do not support the -WhatIf parameter, so it has been removed from the doc. Only updated files under the /teams/teams-ps/teams path to leave the SkypeForBusiness Server docs as is.


sdwheeler:

PR #11735, opened 2024-04-12
Fix a few typos

Blake-Madden:


chrisda:
@Blake-Madden, I merged since the changes were benign, but be careful about making changes across unrelated top level folders in this repo. I monitor Exchange only; others are responsible for monitoring Skype/Teams; I'm not clear of ownership on other folders, but it isn't me.

PR #11738, opened 2024-04-16
fix for scanner statement

cabailey:


cabailey:
#signoff & merging - minor update confirmed by Sarah

PR #11798, opened 2024-05-02
Add SdnRuntimeConfiguration to Set-CsTeamsMeetingBroadcastConfigurati…

musingh1:
…on.md


williamlooney (changes requested):
Can you update the description for this parameter to follow the format of other parameters, something along the lines of : 

"Specifies connection parameters used to connect with a 3rd party eCDN provider. These parameters should be obtained from the SDN provider to be used."

williamlooney (approved):
Approved .

PR #11799, opened 2024-05-02
Adding the missing properties to New-CsTeamsEventsPolicy.md

musingh1:


teams/teams-ps/teams/New-CsTeamsEventsPolicy.md
```diff
@@ -132,7 +147,41 @@ Aliases:
...
 Required: False
 Position: Named
-Default value: Everyone
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -AllowedTownhallTypesForRecordingPublish
+This setting describes how IT admins can control which types of Town Hall attendees can have their recordings published.
```
williamlooney, line 156: This description is not clear. Can you follow the same style as other parameters with something like:

"This setting governs which Town Hall attendees can have their recordings published".

teams/teams-ps/teams/New-CsTeamsEventsPolicy.md
```diff
@@ -132,7 +147,41 @@ Aliases:
...
+
+Required: False
+Position: Named
+Default value: None
+Accept pipeline input: False
+Accept wildcard characters: False
+```
+
+### -AllowedWebinarTypesForRecordingPublish
+This setting describes how IT admins can control which types of webinar attendees can have their recordings published.
```
williamlooney, line 173: This description is not clear. Can you follow the same style as other parameters with something like:

"This setting governs which webinar attendees can have their recordings published".

williamlooney (changes requested):

teams/teams-ps/teams/New-CsTeamsEventsPolicy.md
musingh1: This policy is a bit confusing. I got the text from the corresponding Set cmdlet. https://learn.microsoft.com/en-us/powershell/module/teams/set-csteamseventspolicy?view=teams-ps

Let me follow up with our product team to get the right string. Will update the set policy as well.

PR #11830, opened 2024-05-15
Removed line stating this cmdlet returns Security Groups

paisvivekanand:
This cmdlet returns only mail enabled security groups and does not return all security groups. Made changes accordingly.


chrisda:
@paisvivekanand, the description of the RecipientTypeDetails parameter clarifies what types of groups are returned in on-prem vs. online (and to be clear, this article is single-source for on-prem and online because the cmdlet is available in both environments). 

I modified your description to clarify.

PR #11911, opened 2024-06-10
Update Set-CASMailbox.md deprecating -IsOptimizedForAccessibility paramter

LynnetteZhong:
This parameter will not be supported by OWA team by Aug 19th, 2024. As OWA team will no longer support the light version outlook on the web.

### DIFF
--- a/exchange/exchange-ps/exchange/Set-CASMailbox.md
+++ b/exchange/exchange-ps/exchange/Set-CASMailbox.md
@@ -700,7 +700,7 @@ Accept wildcard characters: False
 ```
 
 ### -IsOptimizedForAccessibility
-**Note**: This paramater will be deprecated from Exchange Online.
+This parameter is functional only in on-premises Exchange.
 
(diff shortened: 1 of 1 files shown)


chrisda:
@LynnetteZhong:

- Syntax issues: you can't add extra text to parameters in the Syntax block or the section heading of the parameter description sections.
- Never a good idea to mention specific dates. They have a nasty habit of sliding.
- Cmdlet reference articles are single-sourced for Exchange Server (2010-2019) and ExO, so unless this thing is also going away from on-prem Exchange Server (doubtful), the parameter needs to remain in the article.
- The parameter is already described as "Functional in on-prem Exchange only", which by definition means it doesn't work in ExO. So, whether it's available but doesn't work vs. it's not available shouldn't confuse customers (either way, it doesn't work for them).
- If/when that parameter is actually unavailable to customers in ExO PowerShell, I'll make the following changes to the parameter description:
  - Remove the value "Exchange Online" from the "Applicable:" YAML property.
  - Change the first line of the description to, "This parameter is available only in on-premises Exchange."

PR #11915, opened 2024-06-14
Update Set-CsTeamsAcsFederationConfiguration.md

jamescadd:
Adds identity parameter

### DIFF
--- a/teams/teams-ps/teams/Set-CsTeamsAcsFederationConfiguration.md
+++ b/teams/teams-ps/teams/Set-CsTeamsAcsFederationConfiguration.md
@@ -80,6 +80,22 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -Identity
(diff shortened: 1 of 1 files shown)


teams/teams-ps/teams/Set-CsTeamsAcsFederationConfiguration.md
```diff
@@ -80,6 +81,22 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -Identity
+Specifies the collection of tenant federation configuration settings to be modified. Because each tenant is limited to a single, global collection of federation settings there is no need include this parameter when calling the Set-CsTenantFederationConfiguration cmdlet. If you do choose to use the Identity parameter you must also include the Tenant parameter. For example:
```
American-Dipper, line 85:
```suggestion
Specifies the collection of tenant federation configuration settings to be modified. Because each tenant is limited to a single, global collection of federation settings there is no need include this parameter when calling the Set-CsTenantFederationConfiguration cmdlet. If you do choose to use the Identity parameter, you must also include the Tenant parameter. For example:
```

PR #11999, opened 2024-07-17
Remove Desks from Set-Place.md

Julia-Foran:
Removing the -Desks parameter which has been deprecated / is not in use.


chrisda:
@Julia-Foran, I'll go ahead and approve, but be advised that parameter is still available in two different M365 E5 tenants as of 7/19. If it's still there in ~Sept/Oct, I reserve the right to add it back with the description, "This parameter is reserved for internal Microsoft use". Availability vs. functionality are two different things, and for sanity's sake in keeping up with 1400+ cmdlets and 14,000+ parameters, I need to keep available parameters in the articles.

Julia-Foran:
@chrisda  - Understood.  Thanks for the clarification about this distinction, too.  I created an ADO item for us to finish removing it in August: https://outlookweb.visualstudio.com/MicrosoftPlaces/_workitems/edit/293412.  I will be on maternity leave at this time, so please check with the assignee on that task if you find it not completed when you check back.

PR #12023, opened 2024-07-23
Update New-CsCallQueue.md

CLYVR:
Updating for launch of Courtesy Callback


mkbond007:
@CLYVR, does the information need to say "public preview" instead of "private preview" today?

CLYVR:
@mkbond007 -ServiceLevelThresholdResponseTimeInSecond is still private preview.

PR #12068, opened 2024-08-13
Update Move-CsUser.md to address minimum MicrosoftTeams module versio…

VikneshMSFT:
…n requirement

We have introduced some changes on cmdlets available in MicrosoftTeams module to deprecate some left out functionality as part of SFBO infra decoupling.


michbrown-png:
@VikneshMSFT - approved

PR #12107, opened 2024-09-05
Update Set-Label.md with content mark alignment clarification

cabirch:


cabirch:
@chrisda This is a small change to clarify that PPT also has the same behavior as Word related to content marking margin offsets

PR #12174, opened 2024-10-15
Update Export-ActivityExplorerData.md

mayanksahu11:
Adding the list of supported filters


chrisda:
@mayanksahu11, these articles are also converted and used for help at the command line (`Get-Help <Cmdlet>`) so the formatting options are very limited. I had to convert your table to a list.

PR #12183, opened 2024-10-18
Patch 11

przlplx:
revising "Applies to" metadata in sweep of Skype for Business content


michbrown-png:
@przlplx , the parameters here still list skype list skype for business online? Can you remove the "applies to Skype for Business Online" from all parameters?
image

PR #12332, opened 2024-10-29
Update Set-CsCallQueue.md

CLYVR:
Updating to remove Private Preview comment on Callback related parameters as Callback is now GA.


CLYVR:
@mkbond007 Please and thank you - missed updating some of the PS parameters for Callback.

CLYVR:
And updating for new option with New-CsAutoAttendant

PR #12377, opened 2024-11-20
Link fixes

chrisda:
And other minor formatting fixes as found


pavellatif:
@chrisda looks like build is failing

chrisda:
@pavellatif, yes. The errors are nonsense; there are no .yml versions of .md article files anywhere in the repo; I suspect it's something with the build process that changed. I have an open item with "Ask an admin."

PR #12421, opened 2024-12-12
Update TeamsCallingPolicy.copilot docs

blackquack:


teams/teams-ps/teams/New-CsTeamsCallingPolicy.md
```diff
@@ -396,8 +396,8 @@ Accept wildcard characters: False
 Setting this parameter lets you control how Copilot is used during calls and if transcription is needed to be turned on and saved after the call.
 
 Valid options are:
-- Enabled: Copilot can work with or without transcription during calls. This is the default value.
-- EnabledWithTranscript: Copilot will only work when transcription is enabled during calls.
+- Enabled: Copilot will only work when transcription is enabled during calls. Transcription will be asked to turned on before proceeding and will be saved after the call. This is the default value.
+- EnabledWithTranscript: Copilot can work with or without transcription during calls. If transcription was not turned on while using Copilot, it will not be saved after the call.
```
michbrown-png, line 400: I am a little confused here. Why does the possible value "Enabled" now mean that copilot only works when Transcription is turned on and "EnabledWithTranscript" means it can work with or without transcription. The previous definitions were:

- Enabled: Copilot can work with or without transcription during calls. This is the default value.
- EnabledWithTranscript: Copilot will only work when transcription is enabled during calls.

michbrown-png (changes requested):

teams/teams-ps/teams/New-CsTeamsCallingPolicy.md
blackquack: There was some changes from PM, we are changing the interaction with the default value. The public docs should reflect this updated behavior

blackquack: updated the defaults from policy definition side, description kept same but marked with updated default value

PR #12509, opened 2025-02-05
Update Set-CsCallQueue.md

CLYVR:
Removing reference to callback parm’s in caution as these are now in TAC.


CLYVR:
@mkbond007 Please and thanks!

mkbond007:
@CLYVR - Don't we still want to keep the PS param info for the persons that still use PS instead of TAC?

CLYVR:
@mkbond007 The caution was that changes in TAC removed the PS configuration as TAC wasn't aware of the callback feature. Now that it's aware this caution no longer applies - changes can be made through TAC or PS and nothing is lost now.

PR #12523, opened 2025-02-14
Update Invoke-ComplianceSecurityFilterAction.md

asharp-m:
Submitted via Learn Editor.


asharp-m:
We were missing a - in front of the action parameter in one of the examples so quickly added a correction for review.

Nina Chen might want to approve or be included for awareness.

PR #12527, opened 2025-02-17
On premise only policies

cibelera:
Remove the reference to Exchange Online to the policies that are on prem only


chrisda:
@cibelera, closing per email description.

PR #12528, opened 2025-02-18
Update Set-Label.md with clarification on applycontentmarkingfooterma…

cabirch:


cabirch:
@chrisda this is a super small wording change (requested in a conversation with a customer)

PR #12532, opened 2025-02-19
Minor edit - Note format

alekyaj:


chrisda:
@alekyaj, any special formatting like Alerts is lost when the markdown files are combined and converted to XML for code check-in for help at the command line (Get-Help <Cmdlet>). They look great on the web, not so great in ASCII text. There was also a period of time where some areas of the cmdlet reference article didn't support Alert formatting (the SYNOPSIS), so I simply don't use it.

alekyaj:
> @alekyaj, any special formatting like Alerts is lost when the markdown files are combined and converted to XML for code check-in for help at the command line (Get-Help ). They look great on the web, not so great in ASCII text. There was also a period of time where some areas of the cmdlet reference article didn't support Alert formatting (the SYNOPSIS), so I simply don't use it.

@chrisda, Noted! Please let me know if you want me to revert the note. I will keep this in mind moving forward. Thanks for the detailed explanation.

chrisda:
@alekyaj, no need to revert. I might change it in the future.

PR #12556, opened 2025-02-28
Update Set-ArcConfig.md to allow removal of final entry

supermistag:
Added clarification on how to remove the last entry (or all entries) from the list, as the command cannot handle an empty array.

Update wording to refer to arc sealers instead of "file types" as this did not make sense.


supermistag:
@microsoft-github-policy-service agree company="Trustwave Holdings, Inc."

PR #12564, opened 2025-03-04
Removing Immersive events TPM Documentation

Erikkre:
Removing Immersive Events setting in TeamsEventsPolicy


michbrown-png:
We are temporarily removing the documentation for Immersive Events because we are under NDA. It will be released on May 19th

PR #12565, opened 2025-03-04
Updated for creating allow directly from TABL

dhairyya:
Can add sender and URL allow from TABL


dhairyya:
@chrisda could you please merge this for me?

PR #12572, opened 2025-03-06
Update New-AutoSensitivityLabelRule.md

tiffanyyumsft:
update docs based on escalation for by design limitation of using edm sits


tiffanyyumsft:
@chrisda  for review and publication

PR #12617, opened 2025-03-21
Update Update-M365TeamsApp.md

HayasakaKenta:
The descriptions for example 2 and example 3 do not match the PowerShell command, so fix that.


sdwheeler:

PR #12659, opened 2025-04-03
Update New-CsTeamsVoiceApplicationsPolicy.md

CLYVR:
Removing comment regarding MWBT feature not being available as it's now GA.


CLYVR:
@mkbond007 PS cmdlet changes for MWBT.  Thanks!

PR #12672, opened 2025-04-08
Update Set-CsPhoneNumberAssignment.md

raul8a07:
The Set-CsPhoneNumberAssignment cmdlet just went live in the DoD around April 6th, 2025. The Set-CsUser cmdlet has been depracated. 

I included GCC-H and DoD since these are the tenants that usually receive updates at the same time.


sdwheeler:

PR #12703, opened 2025-04-29
Update Set-CsTeamsFilesPolicy.md

apoorv33:
Updating a setting which is currently in preview (available only in inner rings), based on the recommendations of Microsoft Security teams. This will be re-updated once the feature rollout is complete.


sdwheeler:

PR #12755, opened 2025-05-17

arpitha-dhanapathi:
Removed IRM policies for now (it will come in June) updated CCSI rule along with description.


arpitha-dhanapathi:
@chrisda - FYIP. Thanks,

PR #12804, opened 2025-06-17
Update Get-EXOMailbox.md

akash-bhargava:
Submitted via Learn Editor.


akash-bhargava:
Removed GroupMailbox from Valid values under RecipientTypeDetails as it is not supported by the cmd-let

PR #12825, opened 2025-06-23
Update Set-DlpCompliancePolicy.md

kyracatwork:
Updated description for enforcementplane support on browser


kyracatwork:
cc @chrisda for awareness & @brenle FYI for future network addition

PR #12827, opened 2025-06-23
Update Set-FeatureConfiguration.md

kyracatwork:
Added enforcementplane for browser and network for collection policies


kyracatwork:
cc @chrisda for awareness & @brenle because I added network, let me know if this shouldn't be the case.

kyracatwork:
Not the right update/removed

PR #12859, opened 2025-07-07
Update SamAccountName description

hfukumotoms:
The explanation "The maximum length is 20 characters." has been removed. 

This explanation applies to user objects (Set-Mailbox), and it is possible to set a SamAccountName longer than 20 characters for group objects. In fact, in Exchange Server 2016 and 2019, you can create groups with a SamAccountName longer than 20 characters using the Set-DistributionGroup. (Set-Mailbox will fail with an error.)


sdwheeler:
The name of the folder has changed. These files are now in `exchange/exchange-ps/ExchangePowerShell`.

PR #12876, opened 2025-07-11
Migrate officewebapps docset to use PSMD2Yaml and new page template

xmdanni:
When `MAML2Yaml` is not in the path of `monikerMapping.json`, reference content in the docset will be built by PSMD2Yaml instead of its predecessor MAML2Yaml. For details, see https://dev.azure.com/ceapex/Engineering/_workitems/edit/783982


xmdanni:
#sign-off

PR #12881, opened 2025-07-14
Update to new PowerShell reference schema

Saisang:
Update to new PowerShell reference schema


Saisang:
@garycentric Could you hep to merge this?

PR #12902, opened 2025-07-18
Remove duplicate text from module file

sdwheeler:
Remove duplicate text from module file

- Fixes AB#454968


sdwheeler:
#sign-off

PR #13037, opened 2025-08-12
Update New-MoveRequest.md to remove BIL and LIL from ExOnline

wesmimsft:
Updating documentation to indicate that BadItemLimit and LargeItemLimit are no longer supported in Exchange Online and refer tenant admins to Data Consistency Score.


chrisda:
Confirmed the parameters aren't available in a plain M365 E5 tenant. Important because a parameter not working vs. not available are two different issues, each requiring a different documentation approach.

PR #13088, opened 2025-09-04
Update Set-LabelPolicy.md

RC22-Dev:
Submitted via Learn Editor.

### DIFF
--- a/exchange/exchange-ps/ExchangePowerShell/Set-LabelPolicy.md
+++ b/exchange/exchange-ps/ExchangePowerShell/Set-LabelPolicy.md
@@ -175,33 +175,6 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
-### -AddExchangeLocationException
-
-> Applicable: Security & Compliance
-
-The AddExchangeLocationException parameter specifies the mailboxes to add to the list of excluded mailboxes when you use the value All for the ExchangeLocation parameter. A valid value is a mailbox.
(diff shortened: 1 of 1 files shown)


chrisda:
@carterryanmsft some points:

1. When you remove a parameter, you can't just remove the parameter section. You also need to remove references to the parameter from the Syntax block.
2. I just checked in a plain M365 E5 tenant (no SDF, no Preview, no Flighting) and that parameter is available. In general, available parameters must remain in the article whether they work or not. If the parameter truly doesn't work, we should change the description to "This parameter is reserved for internal Microsoft use." If the parameter is ever truly made inaccessible, we can remove the parameter from the article.
3. I would presume the RemoveExchangeLocationException parameter is similarly affected?

ghost:
> @carterryanmsft some points:
> 
> 1. When you remove a parameter, you can't just remove the parameter section. You also need to remove references to the parameter from the Syntax block.
> 2. I just checked in a plain M365 E5 tenant (no SDF, no Preview, no Flighting) and that parameter is available. In general, available parameters must remain in the article whether they work or not. If the parameter truly doesn't work, we should change the description to "This parameter is reserved for internal Microsoft use." If the parameter is ever truly made inaccessible, we can remove the parameter from the article.
> 3. I would presume the RemoveExchangeLocationException parameter is similarly affected?

Thanks Chris, new to this. I am confirming with the Feature team.

PR #13133, opened 2025-09-23
Update and rename New-CsTagTemplate.md to New-CsTagsTemplate.md

CLYVR:
Updating from TagTemplate to TagsTemplate to match the PS cmdlet change


CLYVR:
@mkbond007 Please and thanks.

PR #13580, opened 2026-05-13
Mount-Database example identity parameter does not include servername

iserrano76:
Updated the Mount-Database command example 
It does not include the server name

This is how it is in my lab:
image


guruxp (approved):

guruxp:
Just for clarity what Ignacio is trying to do here is removing the server name from the db to be mounted as Exchange only needs the database name

PR #13604, opened 2026-05-19
Update Get-CsCallQueue.md with important notice re Statistics-Current Queue Size Output

CLYVR:
Added notice about upcoming removal of queue size output.


andreitapalaga:
Hi,

We use this line output to get number of calls in queue.
Will this mean that Get-CsCallQueue real time data around number of calls in queue will not be retrieved anymore after 28th of August?
If yes, will there be another option available for partners to get number of calls in queue in real time?

PR #13643, opened 2026-06-04
Document EwsAllowedAppIDs parameter in Set-OrganizationConfig

grtaylor806:
Added EwsAllowedAppIDs parameter documentation for EWS access control. @chrisda


chrisda:
@grtaylor806, when is this thing going to be public?

grtaylor806:
Hey Chris, thanks for the help on this. The code is out already, but gated so it can't be set just yet. We're planning to enable starting next week. So just wanted to get this page updated before then. If we can push this out next week, that would be perfect, thanks.

PR #13743, opened 2026-08-05
Add RecipientDelimiters parameter description

souzalucasr:
## Summary
- Add the missing RecipientDelimiters parameter description for Set-OrganizationConfig.

## Validation
- Confirmed the placeholder was replaced exactly once.
- Ran `git diff --check`.

### DIFF
--- a/exchange/exchange-ps/ExchangePowerShell/Set-OrganizationConfig.md
+++ b/exchange/exchange-ps/ExchangePowerShell/Set-OrganizationConfig.md
@@ -4287,7 +4287,10 @@ Accept wildcard characters: False
 
 This parameter is available only in the cloud-based service.
 
-The RecipientDelimiters parameter specifies whether there is a tenant-wide default value for the Mail Compose and Reply setting that controls whether commas are used to separate recipients.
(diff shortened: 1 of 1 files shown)


chrisda (approved):

souzalucasr:
@chrisda thanks for taking a look and adding clarification here. @emilygrow recommended either removing the last two value/default lines so the PR matches the original requested wording, or moving the “This value is the default” note. I chose the first option and pushed 3df6061 to restore the shorter one-paragraph description on the PR branch.

If there’s a docs convention or Exchange PowerShell requirement we’re missing that requires documenting the `$true` / `$false` behavior here, please let us know. We can adjust the wording instead of removing those lines entirely.

chrisda:
@souzalucasr, I dig the intent, but the one-liner doesn't include the Boolean nature of the parameter ($true/$false required to use the parameter). The majority of Boolean parameter descriptions in Exchange-related cmdlets use that "Valid values are:  • $true: blah • $false: blah" structure. I moved the "This value is the default" statement up to the $true bullet, but we can also remove it if that works better.

souzalucasr:
Thank you, @chrisda, that makes sense. I updated the PR to keep the Boolean valid-values structure and moved the default note to the `$true` bullet, since commas are the default behavior. Let me know if you’d prefer any wording tweaks then. I appreciate your support on this.

PR #13773, opened 2026-08-18
Document BusyOnBusyOption for Get/Set-CsUserCallingSettings

renatogbp:
Documents the per-user Busy on Busy setting: Set-CsUserCallingSettings -BusyOnBusyOption (PlayBusySignal, RedirectAsUnansweredCall, RingUser) and the BusyOnBusyOption output property on Get-CsUserCallingSettings.

### DIFF
--- a/teams/teams-ps/MicrosoftTeams/Set-CsUserCallingSettings.md
+++ b/teams/teams-ps/MicrosoftTeams/Set-CsUserCallingSettings.md
@@ -205,7 +205,7 @@ The busy on busy setting for the specified user. Busy on busy controls what happ
 - `RedirectAsUnansweredCall`: the new incoming call is handled as an unanswered call, that is, it is routed using the user's unanswered call settings (`UnansweredTargetType` and `UnansweredTarget`).
 - `RingUser`: the new incoming call rings the user. This turns busy on busy off for the user.
 
-A per-user setting configured with this parameter applies in addition to the tenant-level and user-level busy on busy configuration set through `Set-CsTeamsCallingPolicy`.
+This setting applies when the user's Teams calling policy has `BusyOnBusyEnabledType` set to `UserOverride`. Otherwise, the policy value takes precedence and this setting is ignored.
 
 If you omit this parameter, the current busy on busy setting for the user is left unchanged.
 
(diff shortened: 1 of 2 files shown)


teams/teams-ps/MicrosoftTeams/Set-CsUserCallingSettings.md
```diff
@@ -168,8 +174,53 @@ Set-CsUserCallingSettings -Identity user7@contoso.com -IsUnansweredEnabled $fals
...
 
+### -BusyOnBusyOption
+
+The busy on busy setting for the specified user. Busy on busy controls what happens to a new incoming call when the specified user is already in a call or in a conference. The supported values are:
+
+- `PlayBusySignal`: the new incoming call is rejected with a busy signal.
+- `RedirectAsUnansweredCall`: the new incoming call is handled as an unanswered call, that is, it is routed using the user's unanswered call settings (`UnansweredTargetType` and `UnansweredTarget`).
+- `RingUser`: the new incoming call rings the user. This turns busy on busy off for the user.
+
+A per-user setting configured with this parameter applies in addition to the tenant-level and user-level busy on busy configuration set through `Set-CsTeamsCallingPolicy`.
```
avinsharma-microsoft: One Concern: The policy takes precedence over this setting.  BusyOnBusyOption  only applies when the user's calling policy has  BusyOnBusyEnabledType  set to  UserOverride . If it's Enabled , Disabled  or  or Unanswered the policy wins and this setting is ignored. Although and the cmdlet doesn't error, so it looks like it worked.
Fix Add/reword: This setting applies when the user's Teams calling policy has  BusyOnBusyEnabledType  set to  UserOverride . Otherwise, the policy value takes precedence.

avinsharma-microsoft (changes requested):
Thanks for this, values and examples look good. Just one correction needed on the BusyOnBusyOption  description, comment inline.

teams/teams-ps/MicrosoftTeams/Set-CsUserCallingSettings.md
renatogbp: Fixed in a701fd2. The parameter description now states that the per-user setting applies only when BusyOnBusyEnabledType is UserOverride; otherwise the policy value takes precedence.

avinsharma-microsoft (approved):
Please don't merge until the version is confirmed. Both files say 8.0.0 or later, but latest on the Gallery is 7.9.0 GA / 7.9.1-preview - 8.0.0 hasn't shipped. Can you confirm it's 8.0.0 and not 7.10.x, and the ship date?

PR #13786, opened 2026-08-26
Update New-CsTeamsAIPolicy, Remove-CsTeamsAIPolicy, Grant-CsTeamsAIPolicy

okadanahoko:
By checking the actual parameters of the Teams Set-CsTeamsAIPolicy command, we confirmed that the parameter used to enable speaker recognition was incorrect.


okadanahoko:
@microsoft-github-policy-service agree company="Microsoft"

